PR #26, opened 2026-02-24
test: fixing the default nf-test

### LINKED ISSUE #15: fix: CI nf-test
### Description of the bug

CI and nf-test were disabled as they failed (or display as passed, but actually fail). 
CI test was removed in the new template, and only rely on nf-test. Fix to comply with this. 
We want working nf-tests at push, using docker and correct nextflow version.

- Update nf-test to run the test profile - a working default nf-test, with an updated snapshot containing software versions, checking files, paths and VCF contents.
- Remove old CI test (also addressed in https://github.com/Clinical-Genomics/oncorefiner/pull/35) 
- Activate nf-test as the new CI test, run using docker and current and latest nextflow versions

### Command used and terminal output

```console

```

### Relevant files

_No response_

### System information

_No response_

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/oncorefiner/issues/15

Fixes the default nf-test, adding snaphot with software versions, variants in the VCFs, paths and file names.

## PR checklist

- [ ] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/oncorefiner/tree/master/.github/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/oncorefiner _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [ ] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [ ] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).


.github/workflows/ci.yml
fellen31, line 1: 👍

.github/workflows/nf-test.yml
```diff
@@ -61,20 +59,17 @@ jobs:
...
-    runs-on: # use self-hosted runners
-      - runs-on=${{ github.run_id }}-nf-test
-      - runner=4cpu-linux-x64
+    runs-on: # use GitHub runners
+      - "ubuntu-latest"
     strategy:
       fail-fast: false
       matrix:
         shard: ${{ fromJson(needs.nf-test-changes.outputs.shard) }}
-        profile: [conda, docker, singularity]
```
fellen31, line 69: 💯

tests/nextflow.config
```diff
@@ -2,13 +2,40 @@
...
+
 */
 
 // TODO nf-core: Specify any additional parameters here
 // Or any resources requirements
 params {
-    modules_testdata_base_path = 'https://raw.githubusercontent.com/Clinical-Genomics/test-datasets/refs/heads/oncorefiner/testdata/'
+    config_profile_name          = 'Test profile'
+    config_profile_description   = 'Minimal test dataset to check pipeline function'
+    modules_testdata_base_path   = 'https://raw.githubusercontent.com/Clinical-Genomics/test-datasets/refs/heads/oncorefiner/testdata/'
```
fellen31: It's good if this is a path to the nf-core modules repository so you can use their test-data. 

```suggestion
    modules_testdata_base_path   = 'https://raw.githubusercontent.com/nf-core/test-datasets/modules/data/'
```

tests/nextflow.config
```diff
@@ -2,13 +2,40 @@
...
     pipelines_testdata_base_path = 'https://raw.githubusercontent.com/Clinical-Genomics/test-datasets/refs/heads/oncorefiner/'
```
fellen31: I would propose this should point to a specific commit rather than the latest commit. 

Otherwise, if you update the test-data, you may break the tests of previous versions of the pipeline.

tests/default.nf.test.snap
```diff
@@ -0,0 +1,98 @@
+{
+    "-profile test": {
+        "content": [
+            {
+                "Workflow": {
```
fellen31, line 36: Why don't we see the software versions from `removeNextflowVersion("$outputDir/pipeline_info/nf_core_oncorefiner_software_mqc_versions.yml")` here? 🤔 

How does `nf_core_oncorefiner_software_mqc_versions.yml` look?

fellen31 (commented):
Awesome!

.github/workflows/ci.yml
beatrizsavinhas, line 1: Could you clarify the reason behind the choice of deleting this test run CI check and only rely on nf-tests instead?
You mentioned "CI test was removed in the new template - look into if we want to keep or not." but there is not information about a decision on this.

.github/workflows/nf-test.yml
beatrizsavinhas, line 1: Sorry, my understanding is limited here... Could you explain why we want to run all the steps in the GitHub runner "ubuntu-latest" and not include the conda profile?

tests/default.nf.test
```diff
@@ -17,6 +17,10 @@ nextflow_pipeline {
             def stable_name = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['pipeline_info/*.{html,json,txt}'])
             // stable_path: All files in ${params.outdir}/ with stable content
             def stable_path = getAllFilesFromDir(params.outdir, ignoreFile: 'tests/.nftignore')
+            // all_vcf_files: All vcf files - can have unstable variants and/or unstable headers
+            def all_vcf_files    = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'])
+            // stable_vcf_files: Vcf files with stable variants - can be snapshot using variantsMD5 - add ignore file if variants are unstable ex ignoreFile: 'tests/.nftignore_unstable_variants'
```
beatrizsavinhas:
```suggestion
            // stable_vcf_files: Vcf files with stable variants - can be snapshot using variants Md5sum values
```
This seems to be a note from during development and not applied in the code.

Also, did you end up figuring out why the Md5sum values are the same even if the run date (included in the header of the file) is different? I am still perplexed about that! 😅

tests/default.nf.test
```diff
@@ -17,6 +17,10 @@ nextflow_pipeline {
...
+            def stable_vcf_files = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'])
```
beatrizsavinhas: This seems to be exactly the same as line 22, which leads me to believe that all vcf files are stable. In that case, I think it makes more sense to just have one variable and provide an explanation in the note.

tests/default.nf.test.snap
beatrizsavinhas: Looks like the `ch_versions` is not being populated in `workflows/oncorefiner.nf` so none of the software versions are actually being captured. 
I'm thinking we should open a separate issue to fix this bug first! Since it's very important that we include in the test the versions of the softwares used.

tests/nextflow.config
beatrizsavinhas: Good point @fellen31! Looks like https://github.com/Clinical-Genomics/test-datasets has not been updated for 4 years?! We should probably not use this repo then.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Initial release of nf-core/oncorefiner, created with the [nf-core](https://nf-co
 - Removed snv_vcf_tbi and sv_vcf_tbi parameter. VCF indexes are now automatically detected [#9](https://github.com/Clinical-Genomics/oncorefiner/pull/9)
 - Renamed pipeline from postprocessing to oncorefiner []()
 - Fixed linting issues [#20](https://github.com/Clinical-Genomics/oncorefiner/pull/20)
+- Fixed nf-test [#26](https://github.com/Clinical-Genomics/oncorefiner/pull/26)
```
beatrizsavinhas: As mentioned above, this PR is fixing/doing several things at the same time. It would be good to elaborate on the description here too.

nextflow.config
```diff
@@ -186,8 +186,8 @@ profiles {
         apptainer.runOptions    = '--nv'
         singularity.runOptions  = '--nv'
     }
-    test      { includeConfig 'conf/test.config'      }
-    test_full { includeConfig 'conf/test_full.config' }
+    test_full   { includeConfig 'conf/test_full.config'      }
+    test        { includeConfig 'conf/test.config'           }
```
beatrizsavinhas: Sure haha! 😆 But is there any reason here for changing the order?

nextflow.config
```diff
@@ -288,3 +288,4 @@ validation {
 
 // Load modules.config for DSL2 module specific options
 includeConfig 'conf/modules.config'
+includeConfig 'conf/modules/prepare_references.config'
```
beatrizsavinhas, line 290: This change seems to be out of scope for this PR. Additionally, also out of scope, if the idea is to have separate configs for each module, what would be the use of the `conf/modules.config`?

beatrizsavinhas (changes requested):
The default test is finally passing, without referring to local file paths!! Much better, well done!!! 🌟 

However, from reading the description of the issue and PR it is not clear to me what changes are intended to be introduced here and what is the reasoning behind these choices. So I have left a lot of comments asking "why?". 😅 
As far as I understand, 2 things are being done at the same time here:
* Update the default nf-test
* Modify the CI checks for nf-tests

Could you please fill in the descriptions and add explanations for the changes? 🙂

tests/default.nf.test.snap
fellen31: Right, makes sense!

tests/default.nf.test
fellen31: This is exactly what `vcf.variantsMD5` from the nft-vcf plugin is for. It only checks the md5sum of the _variants_ (https://github.com/seppinho/nft-vcf?tab=readme-ov-file#variantsmd5). Otherwise, you could just snapshot the file as it is with `snapshot(...).match()`, which snapshots the md5sum of the entire file.

tests/default.nf.test.snap
kristinebilgrav: Oh, I was not aware of this! I can create an issue. Thanks!

tests/default.nf.test
beatrizsavinhas: Ahhh! That makes so much sense! Thank you for explaining @fellen31!!! 🌟

nextflow.config
kristinebilgrav: Haha no, I was playing around with the configs, thats the only reason 😅

tests/default.nf.test
fellen31: The naming of `stable_vcf_files`could probably be better though (comes from Nallo) , e.g. `vcfs_with_stable_variants` or something to make it more self-explanatory.

nextflow.config
kristinebilgrav: Yes, I suppose its out of scope, but it was missed to add here in a previous issue, which was causing some  errors. 
You are right! I guess it can be another issue to fix the configs. I do think that it works for now, but with time we probably want to split them up into module specific configs.

fellen31: > if the idea is to have separate configs for each module, what would be the use of the `conf/modules.config`?

> I do think that it works for now, but with time we probably want to split them up into module specific configs.

I would discourage splitting into one file per module, you would end up with so many files 😱.  Just having `conf/modules.config` could perhaps make sense for very small pipelines, but as they grow that would be an equal nightmare to having one file per module (in my opinion). 

I think having one config per subworkflow and removing `conf/modules.config` is a very reasonable compromise.

tests/default.nf.test
beatrizsavinhas: Yeah, I agree that would be better and would also address [this other comment I left about the `all_vcf_files` and `stable_vcf_files` being essentially the same](https://github.com/Clinical-Genomics/oncorefiner/pull/26#discussion_r2883752281).

nextflow.config
beatrizsavinhas: OK haha!
This is a very small thing so it's OK, but generally it is best to avoid changes like these that don't have a specific purpose.

nextflow.config
beatrizsavinhas: Very good point @fellen31, that I had not considered!

I took a look at the code for `prepare_references` and it seems that it is actually a subworkflow, not a module - `subworkflows/local/prepare_references.nf`. So why is the config under `conf/modules/`? 🤔  Nallo seems to follow this same structure for subworkflow configs.

fellen31: I don't really know, I think it might be a bit of an nf-core relic. It is a config for modules (within a subworkflow), as compared to `conf/base.config`. But, I agree that `conf/subworkflows` would probably make more sense.

.github/workflows/nf-test.yml
kristinebilgrav: We want to run using GitHubs runners -as we are not nf-core and will not use their self-hosted runners. Ubuntu specifies which environment. https://docs.github.com/en/actions/concepts/runners/github-hosted-runners 
For conda, I may remember wrong, but we will only run using docker or singularity, and its therefore unnecessary to test using conda. But let me double check!

tests/nextflow.config
kristinebilgrav: I misunderstood what this was referencing! Indeed it should be the nf-core one. I though this was for our own testdata. Thanks! 🌟

tests/nextflow.config
fellen31: I.e. https://github.com/Clinical-Genomics/test-datasets/tree/1184e1c31b5e47055e3580c7e0f65240a1c005d0/

.github/workflows/nf-test.yml
beatrizsavinhas: Ah alright! Makes sense to use the `ubuntu-latest` runner since we have a linux based system in the server!
Thanks for explaining! 🌟

tests/default.nf.test
kristinebilgrav: All of our vcfs are stable (at least for now), but I noticed that the vcf summary also checkes chromosomes, samples, phasing etc, so thought it would be nice to have also!

tests/default.nf.test
kristinebilgrav: Name updated, thanks!

tests/nextflow.config
kristinebilgrav: This is true. Took some time to figure out how to get that link, but found it - thank you! 🚀

.github/workflows/ci.yml
kristinebilgrav: I have updated the description. This has also been addressed in another PR (https://github.com/Clinical-Genomics/oncorefiner/pull/35)

nextflow.config
kristinebilgrav: Also fixed in (https://github.com/Clinical-Genomics/oncorefiner/pull/36). Changing the filestructure could be something we do when in the future 👀

nextflow.config
beatrizsavinhas: Minor, but I would still suggest restoring this file so we don't run the risk of getting a conflict when updating the template.

tests/nextflow.config
```diff
@@ -2,13 +2,41 @@
 ========================================================================================
     Nextflow config file for running nf-test tests
 ========================================================================================
+ Defines input files and everything required to run a fast and simple pipeline test.
+
+    Use as follows:
+        nf-test test tests/<test> --profile <test,docker>
+        nf-test test --profile=+test,docker
+
```
beatrizsavinhas:
```suggestion
 Defines input files and everything required to run a fast and simple pipeline test.

    Use as follows:
        nf-test test tests/<test> --profile <test,docker>
        nf-test test --profile=+docker

```

The test config is already loaded by default!

But could this information be in documentation instead? It doesn't seem to be standard for other pipelines to include this comment.

tests/default.nf.test.snap
beatrizsavinhas, line 1: So nice to finally have a good snapshot! 🌟

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Removed snv_vcf_tbi and sv_vcf_tbi parameter. VCF indexes are now automatically detected [#9](https://github.com/Clinical-Genomics/oncorefiner/pull/9)
 - Renamed pipeline from postprocessing to oncorefiner []()
 - Fixed linting issues [#20](https://github.com/Clinical-Genomics/oncorefiner/pull/20)
+- Fixed nf-test to run a functional default test, removed old CI test [#26](https://github.com/Clinical-Genomics/oncorefiner/pull/26)
```
```suggestion
- Fixed default test and generate snapshot [#26](https://github.com/Clinical-Genomics/oncorefiner/pull/26)
```
This PR no longer touches on the CI. The changes focus on fixing the default test only!

tests/default.nf.test
beatrizsavinhas: You could also call `.summary` and `.variantsMD5` on the same variable!
But there might be some varying vcf files in the future so I guess you could already have this structure to plan for that.

tests/default.nf.test.snap
```diff
@@ -0,0 +1,119 @@
...
+                "vcfanno/SNV_vcfanno.vcf.gz",
+                "vcfanno/SNV_vcfanno.vcf.gz.tbi",
+                "vep",
+                "vep/SNV_vep.vcf.gz",
+                "vep/SNV_vep.vcf.gz.tbi",
+                "vep/SNV_vep.vcf.gz_summary.html"
+            ],
+            [
+                
+            ],
```
beatrizsavinhas, line 67: Looks like nothing is being outputted for `stable_path`. Is that expected?

beatrizsavinhas (commented):
Nice! 🌟
Strange that there is no output for `stable_path` in the snapshot, though. Could you check that?
Also, I think the title (and if possible descriptions) should be updated. This PR is now fixing the default test only.

tests/default.nf.test.snap
kristinebilgrav: The stable path variable checks the contents of the paths in the outdir, and right now, we dont have any files with stable contents - the MD5 changes since they are VCF files. I have added all files that fail to the .nftignore. So I guess for now it will be empty until we have stable files 😃

tests/nextflow.config
kristinebilgrav: Sure, I dont mind documenting in another way.

tests/default.nf.test
kristinebilgrav: Yes, I think it would be nice to separate them!

tests/default.nf.test.snap
beatrizsavinhas: Alright! Makes sense! Thanks for clarifying. 🙂

beatrizsavinhas (approved):
Looks great! 🌟

PR #36, opened 2026-03-05
fix: prepare references config not being used

fevac:
### Fixed

- Prepare references config not being used


beatrizsavinhas (approved):
Very nice! 🚀 

This bug was also noticed by @kristinebilgrav in https://github.com/Clinical-Genomics/oncorefiner/pull/26 so the fix will also help there!

PR #40, opened 2026-03-10
refactor: Update TEMPLATE branch

beatrizsavinhas:
### Changed

- Updates `.nf-core.yml` according to latest changes in `dev`.
- Updated template with `nf-core pipelines sync`.
- Generate default test snapshot with `nf-test test . --profile=+docker  --update-snapshot`.


.github/workflows/download_pipeline.yml
```diff
@@ -65,7 +57,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install git+https://github.com/nf-core/tools.git@dev
+          pip install git+https://github.com/nf-core/tools.git
```
fevac, line 60: version not fixed?

.github/workflows/ci.yml
fevac, line 1: I'm kinda fond of this CI, specially since I don't seem to be able to run local tests for the time being. What are your thoughts about keeping this file and triggering manually only?

fevac (approved):
Thanks for updating this! 💪 I didn't see anything that I could react to but I'm very new to these templates updates

.github/workflows/ci.yml
beatrizsavinhas: This action would do the same as the default test (run by the nf-test action) except it doesn't even provide a snapshot. So I don't think it's necessary to have both and I guess that's why the template update removes it.
You can simply check the nf-test result that's automatically run or you can also trigger it manually, and you should get what you need!

.github/workflows/download_pipeline.yml
beatrizsavinhas: I'm guessing this will just install the latest nf-core tools version but since this project is no longer tagged as nf-core, it makes sense to use the latest stable version instead of dev? 🤷‍♀️

fellen31: I think — if I remember correctly — that nf-core ran dev versions for their pipelines for a while, but too much broke. So they switch to stable versions in the template, and run dev versions on a few selected pipelines.

beatrizsavinhas: Ahh that makes sense! Thanks @fellen31!

fevac: but we don't want to have it fixed to a specific tag to know which version we run? e.g. 5.3.2?

.github/workflows/ci.yml
fevac: The standard output is a bit different and not as easy to follow, but this is mostly a preference just because I can't get to run local tests nor nf-tests. This should be solved on my side. so nothing really to do here

.github/workflows/download_pipeline.yml
fellen31: This is the way the template is written. _Maybe_ you would like to pin it in this pipeline, but I think it's ok to be always using the latest stable version of nf-core in this check, since it's only checking if the download command is working.

beatrizsavinhas: Yeah I agree. Especially because the nextflow version for the CI that runs the tests is actually fixed: https://github.com/Clinical-Genomics/oncorefiner/blob/70ff34865b5cb5a8430fea2ba008bb445bd1ad87/.github/workflows/nf-test.yml#L21

fellen31: That's the nf-test version, _but_, your idea is correct. It is fixed in e.g. https://github.com/Clinical-Genomics/oncorefiner/blob/70ff34865b5cb5a8430fea2ba008bb445bd1ad87/.github/workflows/linting.yml#L25-L35 with `pip install nf-core==${{ steps.read_yml.outputs['nf_core_version'] }}`.


I'd open an issue/PR in nf-core/tools to change this upstream. I think this is something that would benefit all pipelines. Nice find! ⭐

beatrizsavinhas: Ah of course! `NFT` is nf-test, `NXF` is nextflow 🤦‍♀️ 

But wait, the nf-test action does use specific versions - a set of them:
https://github.com/Clinical-Genomics/oncorefiner/blob/70ff34865b5cb5a8430fea2ba008bb445bd1ad87/.github/workflows/nf-test.yml#L76-L78

Right?

fellen31: The comment was on the version of nf-core tools (which the nf-test workflow doesn't use — lot's of versions here), but this CI could also pin the version of Nextflow. I'd bring it up in nf-core!

beatrizsavinhas: Ah alright, now I'm following! Became indeed confused with all the versions. 😅 
Then you mean that we could fix the nf-core tools version for `download_pipeline` the same way as it's done for `linting`! Thank you!

beatrizsavinhas: Created an issue on nf-core/tools - https://github.com/nf-core/tools/issues/4122.
Thanks again @fellen31!

PR #45, opened 2026-03-10
fix: add_pr_checklist_comment action to run on PRs from forks

beatrizsavinhas:
### Fixed

- Settings for `add_pr_checklist_comment` to allow action to run on a PR originated from a fork:
	- Changed event to `pull_request_target` since this is not subject to any token restrictions.
	- Removed unnecessary token permission settings.


fellen31 (approved):
Bit unsure if we could test this without merging, but let's see if it works!

PR #47, opened 2026-03-11
feat: made all steps optional

fevac:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/43

### Added

- Added new include_process parameter allowing users to selectively enable or disable specific processing steps.

### Changed

- Refactored workflow to use a ch_current_*_vcf pattern, allowing sequential conditional execution of processing steps.
- local annotation, filtering, VEP annotation, and clinical filtering for both SNV and SVs are now conditionally executed based on the include_process configuration.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -17,6 +17,7 @@ params {
     // Temporary inputs
     snv_vcf                     = null
     sv_vcf                      = null
+    include_process             = "snv_local_annotation,snv_filtering,snv_ensemblvep,snv_clinical_filtering,sv_local_annotation,sv_filtering,sv_ensemblvep,sv_clinical_filtering"
 
     // Vep
     vep_cache_version           = 112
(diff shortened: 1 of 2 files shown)


fevac:
For refinement:

- [ ] ch_current_{snv,sv}_vcf  for sequencial steps always including a channel with vcf and index (channel: meta, vcf, tbi)
- [ ] include_process parameter
  - name
  - comma separated
- [ ] supported processes naming accepted by parameter:
  - snv_local_annotation
  - snv_filtering
  - snv_ensemblvep
  - snv_clinical_filtering
  - sv_local_annotation
  - sv_filtering
  - sv_ensemblvep
  - sv_clinical_filtering
- [ ] actual process names
  - VCFANNO
  - RESEARCH_FILTERING
  - ENSEMBLVEP_SNV
  - CLINICAL_FILTERING
  - SVDB_QUERY_DB
  - TABIX_SVDB_SV
  - RESEARCH_FILTERING_SV
  - ENSEMBLVEP_SV
  - CLINICAL_FILTERING_SV

beatrizsavinhas:
Suggestions:

* Avoid overriding and reusing the same input variable for all steps (This is a programming bad practice. For example, if a bug is introduced that makes a mandatory step not run, the pipeline will run anyway, but incorrectly, because the input variable already exists. It also makes it hard to trace back where the contents of the chanel are coming from.)

* Consider changing the logic to focus on `skip` instead of `include`:
	* There can be a `skip_subworkflow` for several connected steps and a `skip_tools` for specific tools instead.

* Consider creating 4 general subworkflows that can be skipped:
  * Local/Ensembl_vep annotation
   * Clinical/Research filtering

  The benefit of subworkflows is that you already have a standardized output by default. No matter what tools are skipped/included in the workflow, the output is always the same `subworkflow.out`.
 
* More general suggestion: Separate SV and SNV processing into 2 subworkflows.

PR #50, opened 2026-03-16
fix: settings for ungrouped parameters

### LINKED ISSUE #27: Fix ungrouped parameters
## Description

Parameters `snv_vcf`, `sv_vcf` and `custom_extra_files` had been added to the code but not to the nextflow_schema. They were added to the `nextflow_schema.json` in https://github.com/Clinical-Genomics/oncorefiner/pull/20 to fix the linting errors, but settings are missing for each parameter and the parameters are now ungrouped.

We need to refactor and add the necessary settings for these parameters.

For `custom_extra_files`, after discussion and comparison to code in nallo https://github.com/Clinical-Genomics/oncorefiner/pull/50#discussion_r2945242496, we have concluded that there is no need to provide sample specific files to VEP and that the extra files necessary for plugins should be provided by the existing `vep_plugin_files` parameter. Parameter `custom_extra_files` should therefore be removed.

## Acceptance criteria

- [x] Description, format, pattern and check if exists is defined and the parameters are included in relevant groups in `nextflow_schema.json`.
- [x] Settings for `snv_vcf`:
  ```
  "snv_vcf": {
      "type": "string",
      "pattern": "^\\S+\\.vcf(\\.gz)?$",
      "description": "<add_description>",
      "format": "file-path",
      "exists": true
  },
  ```

- [x] Settings for `sv_vcf`:
  ```
  "sv_vcf": {
      "type": "string",
      "pattern": "^\\S+\\.vcf(\\.gz)?$",
      "description": "<add_description>",
      "format": "file-path",
      "exists": true
  },
  ```

- [x] Remove unused parameter `custom_extra_files`.

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/oncorefiner/issues/27.

### Fixed

- Added settings and moved parameters `snv_vcf` and `sv_vcf` to `input_output_options` group.
- Updated parameters documentation.


docs/parameters.md
```diff
@@ -10,6 +10,8 @@ Define where the pipeline should find input data and save output data.
 |-----------|-----------|-----------|-----------|-----------|-----------|
 | `input` | Path to comma-separated file containing information about the samples in the experiment. <details><summary>Help</summary><small>You will need to create a design file with information about the samples in your experiment before running the pipeline. Use this parameter to specify its location. It has to be a comma-separated file with 3 columns, and a header row.</small></details>| `string` |  | True |  |
 | `outdir` | The output directory where the results will be saved. You have to use absolute paths to storage on Cloud infrastructure. | `string` |  | True |  |
+| `snv_vcf` | Path to somatic SNV/INDEL VCF file. | `string` |  |  |  |
+| `sv_vcf` | Path to somatic SV VCF file. | `string` |  |  |  |
```
fellen31: From an outsider perspective, it is not clear to me what this means :) Are these paths to files with known somatic variants? Used for what? Filtering?

Or, are these input files per sample, or per case that are to be annotated?

nextflow_schema.json
```diff
@@ -115,6 +131,12 @@
                     "mimetype": "text/csv",
                     "schema": "assets/vep_plugin_files_schema.json"
                 },
+                "vep_custom_extra_files": {
+                    "type": "string",
+                    "description": "Path to extra sample-specific files to be used by VEP for custom annotation.",
```
fellen31: Are these files really specific to each sample? How do you control which files go with which sample? Or do you only run one sample at a time with this pipeline? :)

beatrizsavinhas: These are great questions that I had myself!
The "sample-specific" term I got from looking at the module documentation for that parameter - https://nf-co.re/modules/ensemblvep_vep/.
But then I wanted to see how the parameter was actually used by the module to understand the structure of the parameter (It's named "files" so is it a list? Comma-separated string?). But lo and behold, it doesn't seem to be used at all in the module code?! 🤔 
https://github.com/Clinical-Genomics/oncorefiner/blob/70ff34865b5cb5a8430fea2ba008bb445bd1ad87/modules/nf-core/ensemblvep/vep/main.nf#L11
No usage of "custom_extra_files" or "task.ext.args3".
Am I missing something here?

docs/parameters.md
beatrizsavinhas: Agreed! Can we get some clarity @fevac and @kristinebilgrav? How were you intending that these parameters are used?

nextflow_schema.json
beatrizsavinhas: This is why I suggested to consider removing the parameter in [my previous comment in the issue](https://github.com/Clinical-Genomics/oncorefiner/issues/27#issuecomment-4067824141). Which seems to be what you did in Nallo, right?
E.g:
https://github.com/genomic-medicine-sweden/nallo/blob/4dc94af7f7ad84ccb14fe3ccdfac25279d42de02/subworkflows/local/annotate_svs/main.nf#L41

fellen31: Since VEP has plugins that sometimes needs files as input (SpliceAI for example) they are input through `custom_extra_files`. And yes, since `custom_extra_files` is input together with `vcf` to the module, technically the extra files could be sample specific if the pipeline has the need for it and supports it. E.g. a pipeline with multiple samples of different species running at the same time. 

You need to enable the usage of these files through the config, e.g. adding `--plugin SpliceAI,snv=spliceai_21_scores_raw_snv_-v1.3-.vcf.gz,indel=spliceai_21_scores_raw_snv_-v1.3-.vcf.gz`.

fellen31: In Nallo I input this as a csv/tsv/json/yaml instead of a list: https://github.com/genomic-medicine-sweden/nallo/blob/640eb2c0214fc387aff69ef5a1f212d9e3f0f04a/nextflow_schema.json#L730-L735, since then nf-schema can check that each file exists, has the expected pattern etc. :https://github.com/genomic-medicine-sweden/nallo/blob/dev/assets/schema_vep_plugin_files.json

docs/parameters.md
```diff
@@ -10,6 +10,8 @@ Define where the pipeline should find input data and save output data.
 |-----------|-----------|-----------|-----------|-----------|-----------|
 | `input` | Path to comma-separated file containing information about the samples in the experiment. <details><summary>Help</summary><small>You will need to create a design file with information about the samples in your experiment before running the pipeline. Use this parameter to specify its location. It has to be a comma-separated file with 3 columns, and a header row.</small></details>| `string` |  | True |  |
 | `outdir` | The output directory where the results will be saved. You have to use absolute paths to storage on Cloud infrastructure. | `string` |  | True |  |
+| `snv_vcf` | Path to somatic SNV/INDEL VCF file. | `string` |  |  |  |
```
fevac:
```suggestion
| `snv_vcf` | Path to a VCF file containing somatic SNV/INDEL variants to be processed by the pipeline. The VCF may include one or multiple samples. Typically, this is a single tumor sample for tumor-only analyses, or both tumor and matched normal samples for tumor–normal analyses. | `string` |  |  |  |
```

docs/parameters.md
fevac:
```suggestion
| `sv_vcf` | Path to a VCF file containing somatic structural variants (SVs) to be processed by the pipeline. The VCF may include one or multiple samples. Typically, this is a single tumor sample for tumor-only analyses, or both tumor and matched normal samples for tumor–normal analyses.  | `string` |  |  |  |
```

fevac: fair enough! See if it gets clearer with the suggestion added. I understand you need to add it in the schema file but I wrote it here for clarity

docs/parameters.md
```diff
@@ -33,6 +35,7 @@ Annotation related files and options required for the workflow.
 | `vep_cache_version` | Specify the version of the VEP cache provided to the `--vep_cache` option. | `integer` | 112 |  |  |
 | `vep_cache` | Path to vep's cache directory. <details><summary>Help</summary><small>If no directory path is passed, vcf files will not be annotated by vep.</small></details>| `string` |  |  |  |
 | `vep_plugin_files` | Databases used by both named and custom plugins to annotate variants. <details><summary>Help</summary><small>Path to a file containing the absolute paths to databases and their indices used by VEP's custom and named plugins resources defined within the vcfanno toml file. One line per resource.</small></details>| `string` |  |  |  |
+| `vep_custom_extra_files` | Path to extra sample-specific files to be used by VEP for custom annotation. | `string` |  |  |  |
```
fevac: we use vep for SNVs and SVs and these extra files will be different. So far I don't see any need on having a SV vep custom file. But for clarity we should add this distinction. A suggestion is to rename to `snv_vep_custom_files` or similar

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Updated all modules and removed deprecated `ch_versions` to implement latest nf-core changes that use the `versions` topic channel to collect software versions [#34](https://github.com/Clinical-Genomics/oncorefiner/pull/34)
 - Fixed settings for `add_pr_checklist_comment` to allow action to run on a PR originated from a fork [#45](https://github.com/Clinical-Genomics/oncorefiner/pull/45)
 - Added `species` parameter to provide information for annotation which was previously hardcoded [#49](https://github.com/Clinical-Genomics/oncorefiner/pull/49)
+- Refactored, added settings and moved ungrouped parameters to relevant groups [#50](https://github.com/Clinical-Genomics/oncorefiner/pull/50)
```
fevac:
```suggestion
- Refactored, added settings and moved ungrouped parameters to relevant section [#50](https://github.com/Clinical-Genomics/oncorefiner/pull/50)
```

nextflow_schema.json
fevac: Adding to the conversation: 
- yes, we want to keep this custom files for SNVs. We don't want to running by default now but we want to have the possibility of adding them
- And no, they should not be sample-specific

fevac (changes requested):
Well done! 🌟 Few things to iron but really good comments brought up

docs/parameters.md
fellen31: Then I understand better. It could be more appropriate to input these through the samplesheet. But, the descriptions would still apply, so thanks!

CHANGELOG.md
beatrizsavinhas: I think "groups" makes more sense since they are called parameter groups in the schema.

nextflow_schema.json
beatrizsavinhas: After further discussion with @fellen31 we have cleared out that nallo:
-  Indeed does not use the `custom_extra_files` parameter (see [annotate_svs]([genomic-medicine-sweden/nallo@4dc94af/subworkflows/local/annotate_svs/main.nf#L41](https://github.com/genomic-medicine-sweden/nallo/blob/4dc94af7f7ad84ccb14fe3ccdfac25279d42de02/subworkflows/local/annotate_svs/main.nf#L41)) and [annotate_snvs](https://github.com/genomic-medicine-sweden/nallo/blob/4dc94af7f7ad84ccb14fe3ccdfac25279d42de02/subworkflows/local/annotate_snvs/main.nf#L66)).
- The extra files that are necessary for the plugins are provided by the `extra_files` parameter for VEP. Currently we already have a pipeline parameter `vep_plugin_files` that is provided to VEP's `extra_files` in oncorefiner.

Since @fevac has confirmed that there should not be a need to provide sample specific files, I suggest we follow nallo's approach and only use `vep_plugin_files` but remove the unused parameter `vep_custom_extra_files`.

Do you agree @fevac?

docs/parameters.md
beatrizsavinhas: @fevac Could you clarify one more aspect? If these SNV and SV VCF files are provided, is the samplesheet still required?
Currently, the `input` parameter for the samplesheet is set as required so even if you provide the vcf files to the pipeline, you still have to provide a samplesheet with fastq files.

docs/parameters.md
beatrizsavinhas: The same parameter is currently used as input for both SNV and SV analysis. If we want to provide different files, we should then create separate parameters.

nextflow_schema.json
fevac: Interesting! Sounds good!

docs/parameters.md
fevac: yes, this is an architectural question we have pending to solve. We had discussed whether we wanted to input these as parameters or in the samplesheet. That's why they were considered temporary and were not included in scheme originally. However, it was difficult to decide the best approach without having a better understanding of all input files that the pipeline will need, so we have that discussion blocked for now. 

The current samplesheet does "nothing" and it's currently not relevant at all. It's just there as it came from the nf-core pipeline creation process. It should be either fully removed or updated depending on what we decide during the refinement process.

docs/parameters.md
fevac: Agree but not relevant anymore as this will be removed.

nextflow_schema.json
beatrizsavinhas: Nice! Then I will update the acceptance criteria and create a separate PR to remove the unused parameter!

docs/parameters.md
beatrizsavinhas: Added now!

docs/parameters.md
beatrizsavinhas: Added now!

beatrizsavinhas: Alright! Then I have created an issue to define the input strategy https://github.com/Clinical-Genomics/CancerBioInfo/issues/101 and tried to document the 2 approaches you suggested!

nextflow_schema.json
```diff
@@ -29,6 +29,22 @@
                     "description": "The output directory where the results will be saved. You have to use absolute paths to storage on Cloud infrastructure.",
                     "fa_icon": "fas fa-folder-open"
                 },
+                "snv_vcf": {
+                    "type": "string",
+                    "fa_icon": "fas fa-file",
+                    "description": "Path to a VCF file containing somatic SNV/INDEL variants to be processed by the pipeline. The VCF may include one or multiple samples. Typically, this is a single tumor sample for tumor-only analyses, or both tumor and matched normal samples for tumor\u2013normal analyses.",
+                    "pattern": "^\\S+\\.vcf(\\.gz)?$",
```
fevac:
```suggestion
                    "pattern": "^\\S+\\.vcf\\.gz$",
```
I'm not sure we currently support uncompressed vcf file. I don't think so, at least it hasn't been tested.

nextflow_schema.json
```diff
@@ -29,6 +29,22 @@
...
+                    "format": "file-path",
+                    "exists": true
+                },
+                "sv_vcf": {
+                    "type": "string",
+                    "fa_icon": "fas fa-file",
+                    "description": "Path to a VCF file containing somatic structural variants (SVs) to be processed by the pipeline. The VCF may include one or multiple samples. Typically, this is a single tumor sample for tumor-only analyses, or both tumor and matched normal samples for tumor\u2013normal analyses.",
+                    "pattern": "^\\S+\\.vcf(\\.gz)?$",
```
fevac: same as above

fevac (approved):
Small comment

nextflow_schema.json
beatrizsavinhas: Done too!

PR #49, opened 2026-03-16
fix: Add species parameter

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/oncorefiner/issues/38.

### Fixed

- Added `species` parameter to provide information for annotation which was previously hardcoded.
- Added `species` parameter to `nextflow.config` with default value `homo_sapiens`.
- Updated nextflow schema. Set parameter to enum with `homo_sapiens` as the only option.
- Updated parameters documentation with pre-commit hook.

PR #51, opened 2026-03-17
fix: Remove unused parameter `custom_extra_files`

### LINKED ISSUE #27: Fix ungrouped parameters

beatrizsavinhas:
From https://github.com/Clinical-Genomics/oncorefiner/issues/27.

### Fixed

- Updated parameters documentation.

### Removed

- Unused parameter `custom_extra_files`.

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -19,9 +19,9 @@ Reference genome related files and options required for the workflow.
 
 | Parameter | Description | Type | Default | Required | Hidden |
 |-----------|-----------|-----------|-----------|-----------|-----------|
+| `genome` | Name of the genome reference. (accepted: `GRCh38`\|`GRCh37`) <details><summary>Help</summary><small>Use this parameter to specify the ID for the reference genome used. This is then used to annotate the SV and SNV files e.g. `--genome GRCh38`.</small></details>| `string` | GRCh38 |  |  |
 | `fasta` | Path to FASTA genome file. <details><summary>Help</summary><small>If you don't have a BWA index available this will be generated for you automatically. Combine with `--save_reference` to save BWA index for future runs.</small></details>| `string` |  |  |  |
 | `fai` | Path to FASTA genome index file. <details><summary>Help</summary><small>If none provided, will be generated automatically from the FASTA reference</small></details>| `string` |  |  |  |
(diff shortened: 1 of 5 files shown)


docs/parameters.md
```diff
@@ -19,9 +19,9 @@ Reference genome related files and options required for the workflow.
 
 | Parameter | Description | Type | Default | Required | Hidden |
 |-----------|-----------|-----------|-----------|-----------|-----------|
+| `genome` | Name of the genome reference. (accepted: `GRCh38`\|`GRCh37`) <details><summary>Help</summary><small>Use this parameter to specify the ID for the reference genome used. This is then used to annotate the SV and SNV files e.g. `--genome GRCh38`.</small></details>| `string` | GRCh38 |  |  |
```
beatrizsavinhas, line 22: This change comes from #49 where I forgot to update the parameters document after changing the order of the parameters. 😅

fevac (approved):
Looks good! Just check that this wasn't set on the test profile

beatrizsavinhas:
> Looks good! Just check that this wasn't set on the test profile

Good point! Checked now and the parameter is not used in the test or test config.

PR #35, opened 2026-03-05
fix: Refactor .nf-core.yml and update template

beatrizsavinhas:
### Changed 

- Refactored `genome` parameter to have default value 'GRCh38' and no longer refer to igenomes.

### Fixed

- Refactored .nf-core.yml to:

	- Include updated template settings:
		```
		org: Clinical-Genomics
		...
		skip_features:
	    - igenomes
	    - fastqc
	    ```

	- Removed following checks from list of lint ignore, since these files are no longer expected as a result of the change of organisation:
		```
		files_unchanged:
		    - .github/CONTRIBUTING.md
		    - assets/nf-core-oncorefiner_logo_light.png
		    - docs/images/nf-core-oncorefiner_logo_light.png
		    - docs/images/nf-core-oncorefiner_logo_dark.png
		    - .github/ISSUE_TEMPLATE/config.yml
		 ```

	- Added igenomes and `sendmail_template.txt` files to list of lint ignore, and disable `multiqc_config` checks to fix linting warnings/errors:
		```
		 files_exist:
		    ...
		    - conf/igenomes.config
		    - conf/igenomes_ignored.config
		 files_unchanged:
		    - assets/sendmail_template.txt
		  multiqc_config: false
		```

- Updated template for nf-core/tools version 3.5.2 and merged in order to:
	- Apply changes from settings above. 
	- Apply changes that were missed in https://github.com/Clinical-Genomics/oncorefiner/pull/14.

PR #53, opened 2026-03-17
fix: Update template

beatrizsavinhas:
### Fixed

- Update nf-core tools template.


beatrizsavinhas:
Done in https://github.com/Clinical-Genomics/oncorefiner/pull/35.

PR #54, opened 2026-03-23
fix: `MULTIQC` input

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/105.

### Fixed

- Fixed `MULTIQC` input channel based on solution from [nf-core tools template dev](https://github.com/nf-core/tools/blob/573f105c8e83e3eff7ff0c28e40edc72fa9185a9/nf_core/pipeline-template/workflows/pipeline.nf#L86-L115).
- Updated default test snapshot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Fixed settings for `add_pr_checklist_comment` to allow action to run on a PR originated from a fork [#45](https://github.com/Clinical-Genomics/oncorefiner/pull/45)
 - Added `species` parameter to provide information for annotation which was previously hardcoded [#49](https://github.com/Clinical-Genomics/oncorefiner/pull/49)
 - Added settings and moved ungrouped parameters to relevant groups [#50](https://github.com/Clinical-Genomics/oncorefiner/pull/50)
+- Fixed `MULTIQC` input channel [#54](https://github.com/Clinical-Genomics/oncorefiner/pull/54)
 
 ### `Dependencies`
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Fixed settings for `add_pr_checklist_comment` to allow action to run on a PR originated from a fork [#45](https://github.com/Clinical-Genomics/oncorefiner/pull/45)
 - Added `species` parameter to provide information for annotation which was previously hardcoded [#49](https://github.com/Clinical-Genomics/oncorefiner/pull/49)
 - Added settings and moved ungrouped parameters to relevant groups [#50](https://github.com/Clinical-Genomics/oncorefiner/pull/50)
+- Fixed `MULTIQC` input channel [#54](https://github.com/Clinical-Genomics/oncorefiner/pull/54)
```
fellen31: I've been trying to think about how to best write the changelog vs issue vs PR description to make it clear to the next person what has happened, but I'm not sure what to put were... Not sure this is better.

```suggestion
- Fixed no MultiQC report being produced by the pipeline [#54](https://github.com/Clinical-Genomics/oncorefiner/pull/54)
```

fellen31 (approved):
Nice fix! Cool that you could find it in the dev template.

This looks good to me!

beatrizsavinhas:
> Nice fix! Cool that you could find it in the dev template.

I followed @ramprasadn's directions so I can't take all the credit. 😅

CHANGELOG.md
beatrizsavinhas: Haha I thought of exactly the same when writing this! And I agree with your point, I think we should prioritise providing relevant information to the person reading the log, rather than just describe the changes that were actually made.

Taking your input, I thought of something else and updated it now:

> Fixed bug in `MULTIQC` input channel that prevented the step from running

kristinebilgrav (approved):
Looks good to me! 👍

PR #58, opened 2026-03-27
fix: access parameters in `main.nf` only

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/106

### Changed

- Moved initialisation of input channels to `main.nf`.

### Fixed

- Access parameters in the main workflow only and provide values to the subsequent workflows.
- Bug with input and output for `PREPARE_REFERENCES`:
	- Change name of output variable to reflect the channel variable type.
	- Update type comments.

PR #59, opened 2026-03-27
feat: Add cadd for indels

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/wgs-cancer-pipeline/issues/27

### Added

- CADD score annotation for indels
    - created a subworkflow for the process (annotate_cadd)
        - Only runs if the val_cadd_resources is populated 
    - added modules GAWK, CADD and TABIX
    - compatibility with both GRCh37 and GRCh38 (although only tested with proper data and resources for GRCh38 - see issue for more details)
- test for subworkflow: 
    - NOTE: cadd resources are very large, which makes creating a test dataset hard, but this is being worked on in raredisease https://github.com/nf-core/raredisease/pull/794. We can update the test when this is completed. Therefore: 
        - the subworkflow test is run in `stub` mode only currently    
        - the default test does not run the annotate_cadd subworkflow 

#### Test on hasta
 - The test for GRCh38 was successful (see issue for more details)
 - The test for GRCh37 was not run due to lack of resources.


workflows/oncorefiner.nf
```diff
@@ -62,25 +63,6 @@ workflow ONCOREFINER {
...
-
-        // Gather or get from params
-        ch_vep_cache                 = ( params.vep_cache && params.vep_cache.endsWith("tar.gz") )  ? ch_references.vep_resources
-                                                                                : ( params.vep_cache    ? channel.fromPath(params.vep_cache).collect() : channel.value([]) )
-
-        ch_cadd_header               = channel.fromPath("$projectDir/assets/cadd_to_vcf_header.txt", checkIfExists: true).collect()
-        ch_cadd_resources            = params.cadd_resources                     ? channel.fromPath(params.cadd_resources).map { it -> [[id:'cadd_resources'], it] }.collect()
-                                                                                 : channel.value([])
-        ch_cadd_prescored_indels     = params.cadd_prescored_indels              ? channel.fromPath(params.cadd_prescored_indels).map { it -> [[id:'cadd_prescored_indels'], it] }.collect()
-                                                                                 : channel.value([])
```
kristinebilgrav: Take back

main.nf
```diff
@@ -53,22 +56,31 @@ workflow CLINICALGENOMICS_ONCOREFINER {
                                              : channel.value([[],[]])
 
     PREPARE_REFERENCES (
-        params.vep_cache
+        val_vep_cache
```
kristinebilgrav, line 61: Noticed that this was still param and not val (https://github.com/Clinical-Genomics/oncorefiner/pull/58)

kristinebilgrav: Added as acceptance criteria for https://github.com/Clinical-Genomics/CancerBioInfo/issues/116

subworkflows/local/annotate_cadd/tests/main.nf.test
```diff
@@ -0,0 +1,75 @@
...
+                input[0] = channel.of([
+                    [id:'test'],
+                    file(params.pipelines_testdata_base_path + 'testdata/SNV.tumor.pave.somatic.37.vcf.gz', checkIfExists: true),
+                    file(params.pipelines_testdata_base_path + 'testdata/SNV.tumor.pave.somatic.37.vcf.gz.tbi', checkIfExists: true)
+                    ])
+                input[1] = 'GRCh37'
+                input[2] = channel.fromPath(params.pipelines_testdata_base_path + 'reference/reference.fasta.fai', checkIfExists: true).map {it -> [[id:it.simpleName], it]  }.collect()
+                input[3] = channel.fromPath("$projectDir/assets/cadd_to_vcf_header.txt", checkIfExists: true).collect()
+                input[4] = channel.from("\$PWD").map { dir -> [ [ id: 'cadd_resources' ], dir ] }
+                input[5] = channel.from("\$PWD").map { dir -> [ [ id: 'cadd_prescored_indels' ], dir ] }
```
kristinebilgrav, line 29: Just pointing to a local dir to have some input for the stub test

subworkflows/local/annotate_cadd/main.nf
```diff
@@ -0,0 +1,79 @@
...
+        ch_header                  // channel: [mandatory] [ path(txt) ]
+        ch_cadd_resources          // channel: [mandatory] [ val(meta), path(dir) ]
+        ch_cadd_prescored_indels   // channel: [mandatory] [ val(meta), path(dir) ]
+
+    main:
+
+        ch_rename_chrs_ref    = channel.value([[]])
+
+        // Create files and rename chromosomes if reference is GRCh38
+        if (val_genome.equals('GRCh38')) {
```
kristinebilgrav, line 35: Added an if statement here to make it compatible with both GRCh37 and GRCh38

workflows/oncorefiner.nf
```diff
@@ -33,6 +34,8 @@ include { GENERATE_CYTOSURE_FILES } from '../subworkflows/local/generate_cytosur
 include { paramsSummaryMultiqc   } from '../subworkflows/nf-core/utils_nfcore_pipeline'
 include { softwareVersionsToYAML } from '../subworkflows/nf-core/utils_nfcore_pipeline'
 include { methodsDescriptionText } from '../subworkflows/local/utils_nfcore_oncorefiner_pipeline'
+include { PREPARE_REFERENCES     } from '../subworkflows/local/prepare_references'
```
beatrizsavinhas: Humm this doesn't sound correct. `PREPARE_REFERENCES` is called only by the main workflow. 🤔

workflows/oncorefiner.nf
```diff
@@ -19,6 +19,7 @@ include { SVDB_QUERY as SVDB_QUERY_DB              } from '../modules/nf-core/sv
 include { ENSEMBLVEP_VEP as ENSEMBLVEP_SV          } from '../modules/nf-core/ensemblvep/vep/main'
 include { BCFTOOLS_VIEW as RESEARCH_FILTERING_SV   } from '../modules/nf-core/bcftools/view/main'
 include { BCFTOOLS_VIEW as CLINICAL_FILTERING_SV   } from '../modules/nf-core/bcftools/view/main'
+include { TABIX_TABIX as TABIX_RESEARCH_FILTERING  } from '../modules/nf-core/tabix/tabix/main'
```
beatrizsavinhas: Please add information about this tool and implementation in the PR description.

workflows/oncorefiner.nf
```diff
@@ -46,7 +49,11 @@ workflow ONCOREFINER {
         ch_samplesheet        // channel: [mandatory] samplesheet read in from --input
```
beatrizsavinhas: Check Harshil alignment in this code block, now that new items have been added.

workflows/oncorefiner.nf
```diff
@@ -58,6 +65,7 @@ workflow ONCOREFINER {
         ch_vcfanno_toml       // channel: [optional]  [path(toml_file)]
         ch_vep_cache          // channel: [optional]  [vep_cache_files]
         ch_vep_extra_files    // channel: [optional]  [path(plugin_file1), path(plugin_file2), ...]
+        val_cadd_resources       // string:  [optional]  path to CADD resources directory
```
beatrizsavinhas: Well, I see that we were a bit inconsistent when added these value comments 😅 
Do you think it would be nice to have something like `pat(cadd_resources_dir)` to be consistent with the channels?

workflows/oncorefiner.nf
```diff
@@ -46,7 +49,11 @@ workflow ONCOREFINER {
         ch_samplesheet        // channel: [mandatory] samplesheet read in from --input
         ch_bam_bai_normal     // channel: [optional]  [val(meta), path(bam), path(bai)]
         ch_bam_bai_tumor      // channel: [mandatory]  [val(meta), path(bam), path(bai)]
+        ch_cadd_header           // channel: [mandatory] [ path(txt) ]
+        ch_cadd_prescored_indels // channel: [optional] [ val(meta), path(dir) ]
+        ch_cadd_resources        // channel: [optional] [ val(meta), path(dir) ]
```
beatrizsavinhas:
```suggestion
        ch_cadd_header           // channel: [mandatory] [path(txt)]
        ch_cadd_prescored_indels // channel: [optional] [val(meta), path(dir)]
        ch_cadd_resources        // channel: [optional] [val(meta), path(dir)]
```

workflows/oncorefiner.nf
```diff
@@ -96,14 +104,36 @@ workflow ONCOREFINER {
...
-
-            // VEP
             RESEARCH_FILTERING.out.vcf
                     .map { meta, vcf ->
                         tuple(meta, vcf, [])
                     }
                     .set { ch_vep_snv }
 
+            // ANNOTATE WITH CADD - currently depends on val_cadd_resources - could be improved?
+            if (val_cadd_resources) {
```
beatrizsavinhas: Perhaps this check is not needed? Which would also prevent having this redundant extra input. 
If the tools depend on input channels, if they are empty the process is just not run. You could try to check if you get errors when running the pipeline without the necessary inputs for the subworkflow.

We will have to figure out skip logic in https://github.com/Clinical-Genomics/CancerBioInfo/issues/97 later anyway.

workflows/oncorefiner.nf
```diff
@@ -96,14 +104,36 @@ workflow ONCOREFINER {
...
+
+                TABIX_RESEARCH_FILTERING(RESEARCH_FILTERING.out.vcf) //Subworkflow needs tabix index
+
+                RESEARCH_FILTERING.out.vcf
+                    .join(TABIX_RESEARCH_FILTERING.out.index, failOnMismatch:true, failOnDuplicate:true)
+                    .set{ ch_cadd_in }
```
beatrizsavinhas:
```suggestion
ch_cadd_in = RESEARCH_FILTERING.out.vcf
                    		.join(TABIX_RESEARCH_FILTERING.out.index, failOnMismatch:true, failOnDuplicate:true)
```
(Check indentation in IDE.) According to our fresh [Style guidelines](https://github.com/Clinical-Genomics/oncorefiner/blob/dev/docs/CONTRIBUTING.md#style)! 😆

workflows/oncorefiner.nf
```diff
@@ -96,14 +104,36 @@ workflow ONCOREFINER {
...
             RESEARCH_FILTERING.out.vcf
                     .map { meta, vcf ->
                         tuple(meta, vcf, [])
                     }
                     .set { ch_vep_snv }
 
+            // ANNOTATE WITH CADD - currently depends on val_cadd_resources - could be improved?
+            if (val_cadd_resources) {
+
+                TABIX_RESEARCH_FILTERING(RESEARCH_FILTERING.out.vcf) //Subworkflow needs tabix index
```
beatrizsavinhas: Would it make sense then that this is done in the `ANNOTATE_CADD` subworkflow instead?

workflows/oncorefiner.nf
```diff
@@ -96,14 +104,36 @@ workflow ONCOREFINER {
...
+                    ch_cadd_in,
+                    val_genome,
+                    ch_genome_fai,
+                    ch_cadd_header,
+                    ch_cadd_resources,
+                    ch_cadd_prescored_indels
+                )
+                ANNOTATE_CADD.out.vcf
+                    .join(ANNOTATE_CADD.out.tbi)
+                    .set { ch_vep_snv }
```
beatrizsavinhas:
```suggestion
                ch_vep_snv = ANNOTATE_CADD.out.vcf
                    						.join(ANNOTATE_CADD.out.tbi)
```
Check indentation.

CITATIONS.md
beatrizsavinhas, line 1: Very nice! And well remembered!
However, now we have a few more steps that are needed - see [Adding citations](https://github.com/Clinical-Genomics/oncorefiner?tab=contributing-ov-file#adding-citations).

main.nf
```diff
@@ -92,10 +94,18 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     }
 
     // Reference files
-    ch_genome_fasta         = channel.fromPath(val_genome_fasta).map { it -> [[id:it.simpleName], it] }.collect()
-    ch_genome_fai           = channel.fromPath(val_genome_fai).map { it -> [[id:it.simpleName], it] }.collect()
+    ch_genome_fasta          = channel.fromPath(val_genome_fasta).map { it -> [[id:it.simpleName], it] }.collect()
+    ch_genome_fai            = channel.fromPath(val_genome_fai).map {it -> [[id:it.simpleName], it]  }.collect()
```
beatrizsavinhas:
```suggestion
    ch_genome_fai            = channel.fromPath(val_genome_fai).map { it -> [[id:it.simpleName], it] }.collect()
```

main.nf
```diff
@@ -67,12 +69,12 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
-    ch_sv_vcf               = channel.fromPath(val_sv_vcf).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_sv_vcf_tbi           = channel.fromPath(val_sv_vcf + '.tbi', checkIfExists: true).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_vep_extra_files      = channel.empty()
-    ch_svdb_dbs             = channel.empty()
+    ch_snv_vcf               = channel.fromPath(val_snv_vcf).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
+    ch_snv_vcf_tbi           = channel.fromPath(val_snv_vcf + '.tbi', checkIfExists: true).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
+    ch_sv_vcf                = channel.fromPath(val_sv_vcf).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
+    ch_sv_vcf_tbi            = channel.fromPath(val_sv_vcf + '.tbi', checkIfExists: true).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
+    ch_vep_extra_files       = channel.empty()
+    ch_svdb_dbs              = channel.empty()
```
beatrizsavinhas: What are these being aligned with? 🤔

subworkflows/local/annotate_cadd/tests/main.nf.test
beatrizsavinhas, line 1: I read your note saying that the test would be updated soon, but would it be possible to still take a look at the output directory and file structure? This would help define the config file as well.

beatrizsavinhas: Looks like both `CADD` and `GAWK` have stub settings, so we don't have to run the whole subworkflow with `-stub`, just the two tools. Or does `options "-stub"` apply to all tools that are part of the test run?
In any case, we should still be able to emulate the output file structure.

subworkflows/local/annotate_cadd/tests/main.nf.test
```diff
@@ -0,0 +1,75 @@
...
+        when {
+            params {
+                genome = "GRCh38"
+            }
+            workflow {
+                """
+                input[0] = channel.of([
+                    [id:'test'],
+                    file(params.pipelines_testdata_base_path + 'testdata/SNV.tumor.pave.somatic.37.vcf.gz', checkIfExists: true),
+                    file(params.pipelines_testdata_base_path + 'testdata/SNV.tumor.pave.somatic.37.vcf.gz.tbi', checkIfExists: true)
```
beatrizsavinhas: I think there should be at least a note about updating this test when we start working on [Update test profile to GRCh38](https://github.com/Clinical-Genomics/oncoflow/issues/36).

subworkflows/local/annotate_cadd/tests/nextflow.config
```diff
@@ -0,0 +1,40 @@
...
+        ext.prefix = { "${input.simpleName}_indels" }
+    }
+
+    withName: 'BCFTOOLS_VIEW' {
+        ext.args   = { "--output-type z --types indels,other" }
+        ext.prefix = { "${vcf.simpleName}_indels" }
+    }
+
+    withName: 'CADD' {
+        container = "nf-core/ubuntu:22.04" //Using an basic container because v1.7.3 is too big for CI.
```
beatrizsavinhas, line 14: Nice! 🌟

subworkflows/local/annotate_cadd/main.nf
```diff
@@ -0,0 +1,79 @@
+//
+// A subworkflow to annotate cadd
```
beatrizsavinhas:
```suggestion
// A subworkflow to annotate vcf files with cadd
```
Perhaps this is not the most accurate? 😅 Feel free to correct as you see fit!

conf/subworkflows/annotate_cadd.config
```diff
@@ -0,0 +1,52 @@
...
+        ext.args2 = '\'{original=$1; sub("chr","",$1); print original, $1}\''
+        ext.prefix = "reference_to_cadd"
+        ext.suffix = "txt"
+    }
+
+    withName: '.*:ANNOTATE_CADD:ANNOTATE_INDELS' {
+        ext.args   = { "--columns Chrom,Pos,Ref,Alt,-,CADD --output-type z --write-index=tbi" }
+        ext.prefix = { "${input.simpleName}_ann" }
+        publishDir = [
+            path: { "${params.outdir}/annotations" },
```
beatrizsavinhas, line 62: Ah-ha! So these are the only files we want to keep?

conf/test.config
```diff
@@ -51,4 +51,5 @@ params {
 
     svdb_query_dbs      = params.pipelines_testdata_base_path + 'reference/svdb_querydb_files.csv'
 
+    // TODO make/insert mock input for CADD
```
beatrizsavinhas, line 54: Adding `-stub` to the relevant process(es) should achieve this, no? See my comments below on the subworkflow test.

beatrizsavinhas (changes requested):
I have quite a few comments and suggestions for changes, but for now the most important is to update the branch to the latest version of dev which should change this PR quite a bit.

workflows/oncorefiner.nf
kristinebilgrav: Yes, I was not sure about this either, I think it was something from a dev merge and I was not sure why and how - I´ll fix!

workflows/oncorefiner.nf
kristinebilgrav: Implemented!

workflows/oncorefiner.nf
kristinebilgrav: All channels except for the CADD resources will be populated, therefore I think that most of the subworkflow will run unless we have this, but I will try it out!

subworkflows/local/annotate_cadd/main.nf
kristinebilgrav: Agree! Implemented

workflows/oncorefiner.nf
kristinebilgrav: Agree, moved!

conf/subworkflows/annotate_cadd.config
kristinebilgrav: Yes, exactly! This will be the file that contains the CADD score that will be used for further analysis

subworkflows/local/annotate_cadd/tests/main.nf.test
kristinebilgrav: True! Will add it as a comment in the file and as a criteria in the issue above.

subworkflows/local/annotate_cadd/tests/main.nf.test
kristinebilgrav: Not sure how this works, but I assume that adding `options "-stub"` makes all of it run in stub. If not I think it will run the actual tools? I think that even if we ran some of it not in stub, the only output we want to save is the annotated vcf from the last process. Also not sure how it would be if you mix stub and not when they depend on each other 🤔

conf/test.config
kristinebilgrav: Replied to the comment below. But not sure what you mean here. My intent with this comment is just to remember to add the subsetted dataset for CADD that is being worked on in raredisease so that we can run the subworkflow with the test profile.

main.nf
kristinebilgrav: Tried aligning them to the cadd input below, but maybe does not make sense as there is something inbetween that is not aligned like so?

main.nf
kristinebilgrav: Implemented

CITATIONS.md
kristinebilgrav: Nice - done, thanks!

main.nf
beatrizsavinhas: Yeah, I would say so! 🙂

subworkflows/local/annotate_cadd/tests/main.nf.test
beatrizsavinhas: OK, then even if `options "-stub"` is applied for all steps, there should still be output files produces since that is the stub behaviour defined for each tool. So you should still be able to check if the output directory and file structure are built as intended!

kristinebilgrav: Added all intermediate files to the snapshot as well!

conf/subworkflows/annotate_cadd.config
```diff
@@ -0,0 +1,52 @@
+/*
+Annotate with CADD
+*/
```
beatrizsavinhas, line 12:
```suggestion
//
// Annotate with CADD
//
```
Just to be consistent with the other configs. Perhaps we can skip the other larger comment? 🤔 Let me know what you think!

subworkflows/local/annotate_cadd/tests/main.nf.test.snap
beatrizsavinhas, line 1: Very nice that we can now see all ouput files!!! 🌟

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
```diff
@@ -227,6 +230,7 @@ def toolBibliographyText() {
             bibliography_list +
             vcfanno        +
             bcftools_view  +
+            cadd           +
```
beatrizsavinhas: This is also dependant on `if (val_cadd_resources)`, right? Same above.

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
beatrizsavinhas, line 1: Very nice that you updated this too!

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#60](https://github.com/Clinical-Genomics/oncorefiner/pull/60) Added `GENERATE_CYTOSURE_FILES` subworkflow and necessary nf-core modules `TIDDIT_COV` and `VCF2CYTOSURE`.
 - [#70](https://github.com/Clinical-Genomics/oncorefiner/pull/70) Added `SAMTOOLS/VIEW` for bam to cram conversion in the `main.nf`.
 - [#66](https://github.com/Clinical-Genomics/oncorefiner/pull/66) Added `PROCESS_SNVS` subworkflow.
+- [#59](https://github.com/Clinical-Genomics/oncorefiner/pull/59) Added CADD scoring for InDels in the subworkflow `ANNOTATE_CADD`, with a subworkflow test (stub only)
```
```suggestion
- [#59](https://github.com/Clinical-Genomics/oncorefiner/pull/59) Added `ANNOTATE_CADD` subworkflow, with a test (stub only), for CADD scoring for InDels.
```

The under `Changed` we should also specify that `ANNOTATE_CADD` is now called in `PROCESS_SNVS`!

nextflow.config
```diff
@@ -23,6 +23,11 @@ params {
     bai_normal                  = null
     sex                         = null
 
+    // CADD
+    cadd_resources              = null
+    cadd_prescored_indels       = null
+
+
```
beatrizsavinhas:
```suggestion

```

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -33,37 +33,42 @@ nextflow_workflow {
                     [id:'reference'],
                     file(params.pipelines_testdata_base_path + 'reference/reference.fasta', checkIfExists: true)
                 ])
-                input[1] = channel.of([
+                input[1] = channel.fromPath(params.pipelines_testdata_base_path + 'reference/reference.fasta.fai', checkIfExists: true).map {it -> [[id:it.simpleName], it]  }.collect()
+                input[2] = channel.fromPath("$projectDir/assets/cadd_to_vcf_header.txt", checkIfExists: true).collect()
```
beatrizsavinhas, line 37: I guess it's OK to keep this here for the future, but if the cadd subworkflow is skipped for now this is also not needed, right?

subworkflows/local/annotate_cadd/main.nf
```diff
@@ -0,0 +1,86 @@
...
+
+    main:
+
+        ch_rename_chrs_ref    = channel.value([[]])
+
+        TABIX_INPUT(ch_vcf) //Subworkflow needs tabix index
+
+        ch_vcf
+            .join(TABIX_INPUT.out.index, failOnMismatch:true, failOnDuplicate:true)
+            .set { ch_vcf_tbi }
```
beatrizsavinhas:
```suggestion
        ch_vcf_tbi = ch_vcf
            .join(TABIX_INPUT.out.index, failOnMismatch:true, failOnDuplicate:true)
```
See my previous comments on the new guidelines for defining channels! Same for the code below.

subworkflows/local/annotate_cadd/main.nf
```diff
@@ -0,0 +1,86 @@
...
+            .combine( ch_rename_chrs_ref )
+            .map { meta, vcf, tbi, annotations, annotations_index, header, txt -> tuple( meta, vcf, tbi, annotations, annotations_index, [], header, txt )  } //THERE IS A TBI?
+            .set { ch_annotate }
+
+
+        BCFTOOLS_ANNOTATE_INDELS( ch_annotate )
+
+    emit:
+        vcf = BCFTOOLS_ANNOTATE_INDELS.out.vcf // channel: [ val(meta), path(vcf) ]
+        tbi = BCFTOOLS_ANNOTATE_INDELS.out.tbi // channel: [ val(meta), path(tbi) ]
```
beatrizsavinhas: Super super minor, but should we then start adding spaces between `[  ]`? It's different in the take block above and we haven't done that in other places in the pipeline too, but I see other people are using that notation in their pipelines!

beatrizsavinhas (approved):
Super nice additions to the tests! I left some minor comments and suggestions. But otherwise it looks good! 🌟

conf/subworkflows/annotate_cadd.config
kristinebilgrav: Hmm good point. I think we should add so its consistent, thanks!

subworkflows/local/annotate_cadd/main.nf
kristinebilgrav: I think this might be something copilot does automatically 😅  I dont have a preference - but if people are using that we could change to that as well. But I will remove the spaces for this one so its consistent with what we currently have!

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: This is true! I figured that since the file is there, I might as well set it

beatrizsavinhas: Alright!

CHANGELOG.md
kristinebilgrav: Will add to the comment that its used in `PROCESS_SNVS`

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
kristinebilgrav: Yes! Does this alter how I should write this?

subworkflows/local/annotate_cadd/main.nf
kristinebilgrav: Oh, I missed it - thanks! Fixed the other channels as well.

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
beatrizsavinhas: Yeah, you need to add this if statement otherwise the citation for cadd will show up for all analyses, even if the tool is not run.

kristinebilgrav: Ah, I see!

subworkflows/local/annotate_cadd/tests/main.nf.test
```diff
@@ -0,0 +1,86 @@
...
+
+        then {
+            // All directories and files
+            def output_directories_and_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)
+
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    // All directories and files
+                    output_directories_and_files
```
fellen31: Same comments as in https://github.com/Clinical-Genomics/oncorefiner/pull/73#discussion_r3112543506, I'm not really sure it's appropriate for a subworkflow test to snapshot the output directories, better snapshot the emitted files 😊 

`subworkflows/local/annotate_cadd/tests/nextflow.config‎` might have different output settings not reflecting the pipeline outputs set in `conf/subworkflows/annotate_cadd.config`, so I would say you don't really care about the output structure of this particular test. 

What's important should be to snapshot the emitted vcf (and tbi), e.g.: `{ assert snapshot(workflow.out).match() }`.

beatrizsavinhas: I do see your point! The thing is that most our subworkflows, for now, don't actually emit any output since we're just interested in publishing the files (this might, however, change soon when we [Update publishing strategy to use workflow outputs](https://github.com/Clinical-Genomics/CancerBioInfo/issues/112)). So my thought with this snapshot strategy when I worked on the [`generate_cytosure_files` subworkflow](https://github.com/Clinical-Genomics/oncorefiner/pull/60) was to try to make sure all steps of the subworkflow are actually run, even if only one of the files is emitted according to the _real_ pipeline config.
If we only test the output (besides the fact that it is currently none), how can we be sure that all steps are run in the test? 🤔

fellen31: I think it would be preferable to emit something from the subworkflow, e.g. the processed vcf, even though you don't use it downstream. I don't think it would be wrong, in my opinion it only simplifies when reading the subworkflow when you have a clearly stated input (take) and output (emit) from the subworkflow. And as you say, this will also needed when updating to use workflow output definitions.

If your final output file(s) are created and looks correct (e.g test.cgh, or in this case the cadd-annotated vcf and tbi) I think this should be enough to know that all steps have been run. If you really want to be explicit about it you could do e.g. `assert workflow.trace.succeeded().size() == 3`.

beatrizsavinhas: I do agree that it's a good point to make the code more explicit when it comes to output files! I would be up for implementing the changes you suggest! Thank you for your input @fellen31! 

But in our case right now, this would mean refactoring all the local subworkflow tests we already have. It would be nice to both get input from the rest of the group (I see you're also already talking to @mathiasbio about it!) but also I think it would be easier to refactor everything in one go after we change the output strategy for the whole pipeline. So I would suggest we open an issue to refactor all tests later!
Let's see what the rest say about this!

kristinebilgrav: I agree with the points above! Makes sense to check the output files as they will reflect the subworkflow and if it was run correctly.
I will alter this test back to checking the vcf and tbi.

PR #60, opened 2026-03-30
feat: Add subworkflow `generate_cytosure_files`

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/102.

### Added

- `TIDDIT_COV` from nf-core/modules.
- `VCF2CYTOSURE` from nf-core/modules.

- Subworkflow `GENERATE_CYTOSURE_FILES`.
	- Input:
		- SV VCF file given in the parameters.
		- SV VCF .tbi file.
		- `ch_bam_bai` for tumor and (optionally) normal sample.

	- Steps:
		- `FILTER_VCF` using `BCFTOOLS_VIEW` to filter out `SGL` variants from input `sv_vcf`.
		- `TIDDIT_COV` to generate coverage bed file, for tumor and (optionally) normal sample.
		- `VCF2CYTOSURE` to generate `.cgh` file for tumor and (optionally) normal sample.

	- Config: `conf/subworkflows/generate_cytosure_files.config`.

	- Test with tumor sample.
	- Test snapshot.

### Changed

- Added `ch_bam_bai_normal` and `ch_bam_bai_tumor` to input of `ONCOREFINER` for downstream use by `GENERATE_CYTOSURE_FILES`.
- Updated default test snapshot to include new `.cgh` files.


conf/subworkflows/generate_cytosure_files.config
```diff
@@ -0,0 +1,57 @@
...
+    withName: '.*GENERATE_CYTOSURE_FILES:VCF2CYTOSURE_TUMOR' {
+        ext{
+            args = { [
+                "--bins 20",
+                "--sex ${params.sex}"
+            ].join(' ') }
+            prefix = { "tumor" }
+        }
+        publishDir = [
+            path: { "${params.outdir}/vcf2cytosure" }
```
beatrizsavinhas, line 41: I assumed the `cgh` file was the only file from this subworkflow that we are interested in keeping.

fevac: yes!

subworkflows/local/generate_cytosure_files/main.nf
kristinebilgrav, line 1: Nice!

tests/default.nf.test.snap
```diff
@@ -73,7 +85,8 @@
                 "vep/SNV_vep.vcf.gz_summary.html"
             ],
             [
-                
+                "normal.cgh:md5,df9f0942e27e9c8f576c3d58a2e4ce29",
+                "tumor.cgh:md5,6982664c6f78b094fd4a26afc25e153e"
```
kristinebilgrav, line 89: Dont know how these file look, but will the md5 sum change from time to time here?

beatrizsavinhas: They seem to be html files that contain information about the variants. I presume they shouldn't change since they are directly generated based on the variants present in the input vcf file. The fact that the CI tests ran and succeed already shows that there was no change in the content of the files.
As I mentioned before, I can't properly open the files because the software is proprietary, so I only took a look at the raw html file. But if this causes issues in the future we can always patch these.

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -0,0 +1,47 @@
+nextflow_workflow {
+
+    name "Test Workflow GENERATE_CYTOSURE_FILES"
+    script "../main.nf"
+    workflow "GENERATE_CYTOSURE_FILES"
+    tag "subworkflows"
+    tag "generate_cytosure_files"
+
+    test("Tumor sample") {
```
fevac:
```suggestion
    test("vcf2cytosure - TN data - using somatic SV vcf") {
```
I don't understand the test from that info, is this what you mean?

subworkflows/local/generate_cytosure_files/tests/main.nf.test.snap
```diff
@@ -0,0 +1,15 @@
+{
+    "Tumor sample": {
```
fevac: update if modified above

fevac (approved):
This looks great! Thanks 🌟

tests/default.nf.test.snap
kristinebilgrav: Okay, seems reasonable!

subworkflows/local/generate_cytosure_files/main.nf
beatrizsavinhas: Thanks! 🌟

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: So the name of this set of tests is already stated in line 3 - "Test Workflow GENERATE_CYTOSURE_FILES", which is also displayed when the test is run. I don't see a need to repeat the same information again.

But I think the information about the vcf file is relevant! Perhaps we don't need to add tests for different kinds of vcf files, since the point of this test is just to see if the `.cgh` files are generated. But I can specify in the name that we are using the somatic VCF file from the TN analysis.

subworkflows/local/generate_cytosure_files/tests/main.nf.test.snap
beatrizsavinhas: Done!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: Updated now!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -0,0 +1,52 @@
+nextflow_workflow {
+
+    name "Test Workflow GENERATE_CYTOSURE_FILES"
+    script "../main.nf"
+    workflow "GENERATE_CYTOSURE_FILES"
+    tag "subworkflows"
+    tag "generate_cytosure_files"
+
+    test("Tumor sample using TN data - tumor alignment files and somatic SV vcf files") {
```
kristinebilgrav: A little confused - it says **TN**, but you are only testing tumor only or? And I guess on that topic - do we not want to include a test with a normal produced or is that redundant?

beatrizsavinhas: So the goal of this test is just to see if the `.cgh` file is generated properly.
The subworkflow needs 2 inputs: BAM files from one sample + VCF files from an analysis
Here I'm testing with BAM files from a tumor sample, but using the vcf files from the TN analysis.

I think there's no need to do the same test with the normal sample since it won't test different conditions/logic. This test already shows that the `.cgh` file can be generated and I'm just using the tumor sample as an example.

kristinebilgrav: Okay, I see. I think the `TN data` part of the comment is a little confusing then, but I saw you already discussed it earlier and I guess it does make sense with he VCF being TN. 

Fair!

kristinebilgrav (approved):
Looks great! 🚀

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: I see what you mean, let me try to make it more explicit!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -6,7 +6,7 @@ nextflow_workflow {
     tag "subworkflows"
     tag "generate_cytosure_files"
 
-    test("Tumor sample using TN data - tumor alignment files and somatic SV vcf files") {
+    test("Alignment files from Tumor sample and somatic SV vcf files from TN test data") {
```
kristinebilgrav, line 9: Great, thank you!! 🌟

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -33,13 +33,18 @@ nextflow_workflow {
...
+            def output_files = getAllFilesFromDir(params.outdir)
 
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
-                    output_files
+                    // All directories and files in the output directory
+                    output_directories_and_files,
+                    // All files in the output directory with their md5 checksum
+                    output_files.collect { file -> [ file.getName(), path(file.toString()).md5 ] }
```
fevac, line 47: nice! 🌟

workflows/oncorefiner.nf
```diff
@@ -131,14 +130,14 @@ workflow ONCOREFINER {
         if (ch_sv_vcf) {
 
             // VCF2CYTOSURE
-            ch_bam_bai = ch_bam_bai_normal ? ch_bam_bai_tumor.concat(ch_bam_bai_normal) : ch_bam_bai_tumor
+            ch_bam_bai = channel.empty().mix(ch_bam_bai_tumor, ch_bam_bai_normal)
```
fevac, line 133: aha! nice

fevac (approved):
Looks great! 🌟

workflows/oncorefiner.nf
beatrizsavinhas: Yes, if statements can't really be applied to channels - they always return true, so the previous code wasn't doing anything.

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: You're very welcome!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: Done now!

PR #61, opened 2026-03-30
feat: Add bam and bai params and logic to create joint channels

beatrizsavinhas:
### Added

- Pipeline parameters:
	- `bam_tumor`
	- `bai_tumor`
	- `bam_normal`
	- `bai_normal`

- Logic to create joint channels `bam_bai_normal` and `bam_bai_tumor` (for downstream use - see https://github.com/Clinical-Genomics/CancerBioInfo/issues/102 for further information) in `main.nf`.

### Changed

- Update default test config to include new parameters.


conf/test.config
```diff
@@ -31,6 +31,10 @@ params {
     input               = params.pipelines_testdata_base_path + 'testdata/samplesheet_test.csv'
     snv_vcf             = params.pipelines_testdata_base_path + 'testdata/SNV.tumor.pave.somatic.37.vcf.gz'
     sv_vcf              = params.pipelines_testdata_base_path + 'testdata/SV.tumor.esvee.somatic.37.vcf.gz'
+    bam_normal          = params.pipelines_testdata_base_path + 'testdata/subject_a.normal.redux.bam'
+    bai_normal          = params.pipelines_testdata_base_path + 'testdata/subject_a.normal.redux.bam.bai'
+    bam_tumor           = params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam'
+    bai_tumor           = params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam.bai'
```
kristinebilgrav, line 37: Did you already add the files to the repo? If so you probably need to update `pipelines_testdata_base_path` to that commit.

kristinebilgrav, line 37: Another idea is to do similar as for the tbi files; automatically detect them? They will always be located in the same path as the BAM and it would reduce input params and vals

beatrizsavinhas: I did! But actually, I have found a bug there haha: the `pipelines_testdata_base_path` in the `test.config` (which is what is taken by default for tests) is not specific for a commit; the commit was added instead to `test/nextflow.config` which is currently not used at all. This is why the test are already passing for this PR.

beatrizsavinhas: I prefer to explicitly declare all necessary parameters. It makes it much easier to identify what the pipeline really needs and reduces the risk of a user not having the input directory set up the same way and being surprised by missing files that were never communicated as necessary. This will also make our work a lot easier when we need to transition to a pipeline chain!

kristinebilgrav: Ah, I see! It is used when running subwf tests I think (at least when Ive been working). We probably want the one in `nextflow.config` to be commit specific as well (or just have the parameter once?). Do you mind adding that in this PR?

kristinebilgrav: Hmm, I dont fully agree, but I think this is preference and some of this may change in the future when we refine how we want to set up the samplesheet etc. So if you prefer it, I dont mind!

beatrizsavinhas: Humm yeah, I think that if it's to always be used, it should be added only once (and we have to make sure the correct configs are used for tests). But no, I don't think that should be added to this PR since it is out of scope. I can open a bug report so the issue is documented and can be patched later.

beatrizsavinhas: See https://github.com/Clinical-Genomics/oncorefiner/issues/63.

fevac: Agree with Kristine. index files should be automatically detected and not mandatory parameters. That's the standard bioinfo practice. It's nice to include a check to make sure they are actually present and otherwise  generate the index, but that could be done separately

kristinebilgrav: Great! Thank you ⭐

beatrizsavinhas: I still don't think that fetching the index file from manipulating the directory of the main file with no prior information for the user is a good solution due to the danger I mentioned above. In that case, I think it makes a lot more sense that the parameter is the directory and that we include in the description that the directory should contain both file and index.
I see also that many other pipelines do have parameters for index files (`--fai` seems quite common, for example) so I don't think that logic is standard.

If you both strongly prefer to keep that logic, I will change this to work as the vcf parameters currently do, of course, because the most important is moving forward and we can always re-think it in the future.

fellen31: My two cents is that while yes, it is convenient that many cli bioinformatics tools automatically detect an index when using them in the command line, I'd argue that this pipeline is not meant to be used that way. Our pipelines are configured by cg, or in the case of this pipeline, probably another pipeline. Then I think automatically looking for an index just adds ambiguity, with very little to gain. When there are already established methods for checking if files exist in Nextflow via the schema, I also think it's messier to input a directory and check or hope that files exist in there. 

And just because it's a standard practice, I would not say it's always a good practice. We have a tool in Nallo that modifies and updates the reference directory (!) on Hasta if indexes are missing, which I very much don't like 😄

kristinebilgrav: Hmm. Seems like this should be a part of our refinement of how we want to configure Oncoflow and the running of Oncorefiner, so I think that we can keep it like it is and then note these points for that?

fevac: Agree. Let's keep it like this for now then and we can discuss at a later point the general strategy. Whatever we decide to do, we would need to be consistent (e.g. vcf.gz.tbi behaviour) and update consequently

nextflow.config
```diff
@@ -17,6 +17,10 @@ params {
     // Temporary inputs
     snv_vcf                     = null
     sv_vcf                      = null
+    bam_tumor                   = null
+    bai_tumor                   = null
+    bam_normal                  = null
+    bai_normal                  = null
```
fevac, line 23: Nothing to do at the moment but a thought to keep in mind. In the future, we might want to support either cram or bam.

fevac (approved):
Nice! 🌟

nextflow.config
beatrizsavinhas: Ah good point! We'll have to adjust this in the future.

conf/test.config
beatrizsavinhas: Thank you all for your input! Will keep the parameters here then.
And then work on an issue to add a `tbi` parameter to make it consistent across the pipeline.

PR #66, opened 2026-04-02
feat: Add `PROCESS_SNVS` subworkflow

beatrizsavinhas:
### Added

- Subworkflow `PROCESS_SNVS`.
- Subworkflow config under `conf/subworkflows/process_snvs.config`.
- Subworkflow test with tumor normal data to:
	- Check output files and directories.
	- Check md5sums of stable vcf files.

### Changed

- Moved logic for processing SNV VCF files to `PROCESS_SNVS` subworkflow.
- Moved process settings from `conf/modules.config` to new `conf/subworkflows/process_snvs.config`.
- Updated default pipeline test snapshot to reflect changes in process names.

### Removed

- Removed redundant if statement for `ch_snv_vcf`.


workflows/oncorefiner.nf
```diff
@@ -68,63 +70,21 @@ workflow ONCOREFINER {
         ch_versions             = channel.empty()
         ch_multiqc_files        = channel.empty()
 
-/*
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    ANNOTATE SNVs
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-*/
         // Process SNV VCF files
-        if (ch_snv_vcf) {
```
beatrizsavinhas, line 77: Removed this redundant if statement as we disscussed previously!

tests/default.nf.test.snap
beatrizsavinhas, line 1: Updated default pipeline test snapshot to reflect the new process names.

CHANGELOG.md
```diff
@@ -31,6 +32,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Updated minimum Nextflow version to 25.10.0 [#30](https://github.com/Clinical-Genomics/oncorefiner/pull/30)
 - Added wgs-cancer-pipeline projects list in the issue templates [#37](https://github.com/Clinical-Genomics/oncorefiner/pull/37)
 - Updated link to Contributing Guidelines in the PR checklist to point to the rendered version of the document in `dev` [#56](https://github.com/Clinical-Genomics/oncorefiner/pull/56)
+- Moved logic for processing SNV VCF files to `PROCESS_SNVS` subworkflow [#66](https://github.com/Clinical-Genomics/oncorefiner/pull/66)
```
fevac: what logic does this refer to?

fevac (approved):
🚀

CHANGELOG.md
beatrizsavinhas: Logic that was in `workflows/oncorefiner.nf`. I've updated the comment! Could you check if you agree? Otherwise feel free to modify it as you see fit.

tests/default.nf.test.snap
```diff
@@ -2,28 +2,28 @@
...
                     "bcftools": 1.22
                 },
-                "ENSEMBLVEP_SNV": {
+                "ENSEMBLVEP_SV": {
                     "ensemblvep": 115.2,
                     "perl-math-cdf": 0.1,
                     "tabix": 1.21
                 },
-                "ENSEMBLVEP_SV": {
+                "ENSEMBLVEP_VEP": {
```
kristinebilgrav, line 19: Realize that this naming might not be optimal when we also change the SV part of it? But will probably only be in this file that it will be a bit unintuitive as the settings are in different config files?

kristinebilgrav (approved):
🔥

tests/default.nf.test.snap
beatrizsavinhas: I also thought about how it would look like when we add the SV subworkflow - is the tool going to appear twice? But in any case, this is just a list of tool versions, since it's the same tool, from the same module, so the version should be the same.

kristinebilgrav: Yes, so probably will not be much to worry about!

CHANGELOG.md
beatrizsavinhas: Resolving this discussion as the PR was now approved again.

PR #69, opened 2026-04-07
feat: Add Tumor only and Tumor Normal pipeline tests

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/118.

### Added

- `tumor_only` test config (which includes `test_base.config`) and profile.
- `tumor_only.nf.test` and snapshot.
- `test_tumor_normal.config` (which includes `test_base.config`).

### Changed

- Renamed default test to `tumor_normal.nf.test` and updated snapshot.
- Added default test to lint ignore to prevent linting errors.
- Renamed test config to `test_base.config` and updated content to include common parameters and files for all tests.
- Updated `test` profile to include `test_tumor_normal.config`.
- Changed `.nf-core.yml` to ignore linting checks for:
	- `conf/test.config`
    - `tests/default.config`
    - `tests/default.nf.test`
- Update documentation instructions on running the test suite and pipeline test runs.


tests/tumor_only.nf.test
```diff
@@ -0,0 +1,43 @@
...
+    script "../main.nf"
+    tag "pipeline"
+
+    test("Tumor only analysis with '-profile test'") {
+
+        when {
+            params {
+                outdir    = "$outputDir"
+                bam_normal = null
+                bai_normal = null
```
fevac: vcf files should also be different for tumor only test. TN vcf files will have information for two samples (two columns), while TO only one. We could modify the current vcf files to mimic TO mode. However it would be best to to re-run the test data in tumor only mode to make sure we test with real OA output files (there are differences in how OA runs when in TO vs TN). 

I would argue for having a dedicated profile for TO separated from TN

beatrizsavinhas: Yeah, I had also thought of doing separate TO and TN configs but then I also thought it would be nice to have a default test config so you can run the usual commands `nextflow run -profile test` etc.
Perhaps we could just include the TN config in `conf/test.config`? What do you think of that solution?

beatrizsavinhas: I also agree that it's best to re-run the test data through oncoanalyser in tumor only mode!

fevac: > Yeah, I had also thought of doing separate TO and TN configs but then I also thought it would be nice to have a default test config so you can run the usual commands `nextflow run -profile test` etc. Perhaps we could just include the TN config in `conf/test.config`? What do you think of that solution?

Yes, I like that. Also TN should be the "standard/default", and TO the "special case"

tests/tumor_only.nf.test
```diff
@@ -0,0 +1,45 @@
+nextflow_pipeline {
+
+    name "Test pipeline ONCOREFINER"
+    script "../main.nf"
+    tag "pipeline"
+
+    test("Tumor only analysis with '-profile test'") {
```
kristinebilgrav: Just wondering since it says 'profile test'. Will this also run when running; `nextflow run main -profile test`?

conf/test.config
```diff
@@ -20,13 +20,8 @@ process {
 
 
 params {
-    config_profile_name        = 'Test profile'
-    config_profile_description = 'Minimal test dataset to check pipeline function'
-
-    // Input data
-    // TODO nf-core: Specify the paths to your test data on nf-core/test-datasets
-    // TODO nf-core: Give any required params for the test so that command line flags are not needed
-
```
beatrizsavinhas, line 29: Thought it was time to remove these! 😆

tests/tumor_only.nf.test
beatrizsavinhas: Yes and no, but maybe I'm misunderstanding your question.
No, this is an nf-test so it only runs with the `nf-test` command.
However, this test is analogous to performing a pipeline test run by running `nextflow run . -profile test`, so yes?

tests/tumor_only.nf.test
beatrizsavinhas: Update and clarification: With the current solution we have 2 test profiles (excluding the `test_full` which is not currently used) - `test` and `test_tumor_only`.

* Profile `test` includes the `test_tumor_normal.config`, which should be the "standard/default" as mentioned above.
* Profile `test_tumor_only` includes the `test_tumor_only.config`, which itself includes the other config but modifies only the parameters that are relevant for a TO analysis.

There are also 2 corresponding pipeline tests that are set up so that when running `nf-test` the correct profile is automatically used.

With these 2 profiles it is possible to do pipeline test run for the 2 conditions by running:

* `nextflow run . -profile test` for TN analysis.
* `nextflow run . -profile test_tumor_only` for TO analysis.

docs/CONTRIBUTING.md
beatrizsavinhas, line 1: I thought this would be relevant to update now but let me know if you'd prefer a separate PR for this or if you think additional changes are needed.

nextflow.config
```diff
@@ -191,8 +191,9 @@ profiles {
         apptainer.runOptions    = '--nv'
         singularity.runOptions  = '--nv'
     }
-    test      { includeConfig 'conf/test.config'      }
-    test_full { includeConfig 'conf/test_full.config' }
+    test            { includeConfig 'conf/test_tumor_normal.config' }
+    test_tumor_only { includeConfig 'conf/test_tumor_only.config'   }
+    test_full       { includeConfig 'conf/test_full.config'         }
```
kristinebilgrav, line 200: Out of scope for this PR maybe, but I think we are getting close to being able to remove the test-full part?

kristinebilgrav (approved):
Looks great to me - great work! 🌟

docs/CONTRIBUTING.md
kristinebilgrav: I think you can keep it

nextflow.config
beatrizsavinhas: Fully agree! And we already have an issue for it so it will probably be addressed soon.

fevac (approved):
Very nice! 🌟

conf/test_tumor_only.config
```diff
@@ -0,0 +1,24 @@
...
+includeConfig 'test_base.config'
+
+params {
+    config_profile_name        = 'Test profile - Tumor Normal testdata'
+    config_profile_description = 'Minimal Tumor Normal test dataset to check pipeline function'
+
+    snv_vcf   = params.pipelines_testdata_base_path + 'testdata/tumor_only/subject_a.tumor.purple.somatic.vcf.gz'
+    sv_vcf    = params.pipelines_testdata_base_path + 'testdata/tumor_only/subject_a.tumor.purple.sv.vcf.gz'
+    bam_tumor = params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam'
+    bai_tumor = params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam.bai'
```
kristinebilgrav, line 23: If these files are the same for TN and TO these parameters could be in the common instead of here and above? Or do you think that makes it less tracable?

beatrizsavinhas: Yeah, this is what I was talking about in terms of code duplication for the files for the tumor sample... But since the test data shouldn't change much at all in the lifetime of the pipeline, it shouldn't be an issue. And I think having it defined in each config files makes much more clear which files are actually being used for each run!

kristinebilgrav: Ah I see! I agree, easier to trace this way.

kristinebilgrav (approved):
Well done! 🚀

fevac (approved):
Nice! 🌟 Much cleaner solution, thanks!

PR #70, opened 2026-04-07
feat: add bam-to-cram conversion

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/95

### Added

- samtools view for conversion from BAM to CRAM file format; added to the main.nf after `ONCOREFINER` as this is only for storage and not linked to any other parts of the pipeline. 

### Updated

- updated snapshot


main.nf
```diff
@@ -56,23 +57,12 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
-    // WORKFLOW: Run pipeline
-    //
-
-    // Input channels
-    ch_snv_vcf              = channel.fromPath(val_snv_vcf).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_snv_vcf_tbi          = channel.fromPath(val_snv_vcf + '.tbi', checkIfExists: true).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_sv_vcf               = channel.fromPath(val_sv_vcf).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_sv_vcf_tbi           = channel.fromPath(val_sv_vcf + '.tbi', checkIfExists: true).map { vcf -> [[id:vcf.simpleName], vcf] }.collect()
-    ch_vep_extra_files      = channel.empty()
-    ch_svdb_dbs             = channel.empty()
```
kristinebilgrav, line 73: Not sure why it shows up as if i deleted it? I only moved some channels up

fevac: Don't worry about that. This is how github recognises the changes: they appear as deleted in the original and additions to the new one

main.nf
```diff
@@ -56,23 +57,12 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
+    // Reference files
+    ch_genome_fasta         = channel.fromPath(val_genome_fasta).map { it -> [[id:it.simpleName], it] }.collect()
+    ch_genome_fai           = channel.fromPath(val_genome_fai).map { it -> [[id:it.simpleName], it] }.collect()
+    ch_genome_fasta_fai     = ch_genome_fasta.join(ch_genome_fai, failOnMismatch: true, failOnDuplicate: true)
 
-    // Input for GENERATE_CYTOSURE_FILES
+    // BAM files, input for bam-to-cram and GENERATE_CYTOSURE_FILES
```
beatrizsavinhas:
```suggestion
    // Alignment files
```
Since these are now for more general use, I think it makes sense to generalise the comment.
It could be nice to even join these with the reference files above and just have `// Input files`.

conf/modules/prepare_references.config
```diff
@@ -17,14 +17,21 @@
 
 process {
 
-    withName: '.*:PREPARE_REFERENCES:.*' {
+    withName: '.*PREPARE_REFERENCES:SAMTOOLS_VIEW.*' {
+        ext.prefix = { "${input.simpleName}_${meta.type}" }
+        ext.args =  { '--output-fmt cram'}
         publishDir = [
-            enabled: false,
+            path: { "${params.outdir}/alignments" },
```
beatrizsavinhas:
```suggestion
            path: { "${params.outdir}/alignment" },
```

I think `alignment` is more standardly used.

conf/modules/prepare_references.config
```diff
@@ -17,14 +17,21 @@
...
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
```
beatrizsavinhas:
```suggestion
```

The module already uses the version topic so no `versions.yml` should be produced!

subworkflows/local/prepare_references.nf
beatrizsavinhas, line 1: The subworkflow needs a test! See [Add tests for PREPARE_REFERENCES subworkflow](https://github.com/Clinical-Genomics/CancerBioInfo/issues/116). Since this PR adds complexity to the subworkflow I would argue it makes sense to do it now. If you feel it's out of scope for this PR, consider adding tests first in a separate PR and then just updating them here.

subworkflows/local/prepare_references.nf
```diff
@@ -2,15 +2,42 @@
...
+        //
+        if (ch_bam_bai_normal) {
+            SAMTOOLS_VIEW_NML ( ch_bam_bai_normal, ch_genome_fasta_fai, [[], []], [[],[]], 'crai' )
+
+            SAMTOOLS_VIEW_NML.out.cram
+                .join(SAMTOOLS_VIEW_NML.out.crai)
+                .set { ch_cram_crai_normal }
+        }
+
+        if (ch_bam_bai_tumor) {
```
beatrizsavinhas: There should always be a tumor bam file, no?

subworkflows/local/prepare_references.nf
```diff
@@ -2,15 +2,42 @@
 // Prepare reference files
 //
 
-include { UNTAR as UNTAR_VEP_CACHE                           } from '../../modules/nf-core/untar/main'
+include { UNTAR as UNTAR_VEP_CACHE           } from '../../modules/nf-core/untar/main'
+include { SAMTOOLS_VIEW as SAMTOOLS_VIEW_NML } from '../../modules/nf-core/samtools/view/main'
+include { SAMTOOLS_VIEW as SAMTOOLS_VIEW_TMR } from '../../modules/nf-core/samtools/view/main'
```
beatrizsavinhas: To avoid double imports, tool renaming and code duplication under "Convert BAM to CRAM", I would suggest generalising the subworkflow and provide all available bam files in one input channel to have 2 processes running in parallel in the case of TN.
See example from the `add-vcf2cytosure` branch:
https://github.com/Clinical-Genomics/oncorefiner/blob/77638dbc9ba9b71d29d1a5e712644ecf0f8e8fa5/workflows/oncorefiner.nf#L133-L142

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - Added Nextflow strict syntax compatibility [#30](https://github.com/Clinical-Genomics/oncorefiner/pull/30)
 - Added bam and bai parameters for tumor and normal samples, and logic to create joint bam_bai channels [#61](https://github.com/Clinical-Genomics/oncorefiner/pull/61)
 - Added `sex` parameter [#62](https://github.com/Clinical-Genomics/oncorefiner/pull/62)
+- Added samtools/view for bam to cram conversion [#70](https://github.com/Clinical-Genomics/oncorefiner/pull/70)
```
beatrizsavinhas: Add info on which subworkflow was modified.

beatrizsavinhas (changes requested):
Nice! 🌟 But I have a few comments. The biggest changes I think should be addressed are:
* Adding subworkflow tests. They were already lacking but this PR adds more complexity to the subworkflow.
* Avoiding code duplication in `PREPARE_REFERENCES`.

Let me know if something is not clear enough. I'm happy to explain!

subworkflows/local/prepare_references.nf
```diff
@@ -29,5 +56,7 @@ workflow PREPARE_REFERENCES {
         }
 
     emit:
-        vep_resources = ch_vep_resources // channel: [vep_cache_files]
+        vep_resources = ch_vep_resources    // channel: [vep_cache_files]
+        ch_cram_crai_normal                 // channel: [cram, crai] for normal sample
+        ch_cram_crai_tumor                  // channel: [cram, crai] for tumor sample
```
beatrizsavinhas: Are these files to be used by other tools/subworkflows in the future? If not, no need to emit them.

tests/.nftignore
```diff
@@ -17,3 +17,4 @@ svdb/*.{vcf,vcf.gz,tbi}
 vcfanno/*.{vcf.gz,tbi}
 research/*.{vcf,vcf.gz,tbi}
 research_filtering/*.{vcf,vcf.gz,tbi}
+alignments/*.{cram,crai}
```
beatrizsavinhas, line 20: Do these files need to be ignored? Isn't the content always the same given the bam files in the test data?

subworkflows/local/prepare_references.nf
fevac: Agree. Please make it general

subworkflows/local/prepare_references.nf
```diff
@@ -2,15 +2,42 @@
...
+include { UNTAR as UNTAR_VEP_CACHE           } from '../../modules/nf-core/untar/main'
+include { SAMTOOLS_VIEW as SAMTOOLS_VIEW_NML } from '../../modules/nf-core/samtools/view/main'
+include { SAMTOOLS_VIEW as SAMTOOLS_VIEW_TMR } from '../../modules/nf-core/samtools/view/main'
 
 
 workflow PREPARE_REFERENCES {
     take:
-        val_vep_cache // string: [mandatory] path(cache)
+        ch_bam_bai_normal     // channel: [optional]  channel containing BAM and BAI file for normal sample
+        ch_bam_bai_tumor      // channel: [optional]  channel containing BAM and BAI file for tumor sample
```
fevac: This is not part of the prepare reference workflow. The bam files are not reference files but actual input data. This should be as part of it's own subworkflow or in the main workflow

subworkflows/local/prepare_references.nf
fevac: I'm not sure how the current bam input parameter was set up (maybe it was mandatory for tumor?). But ideally it would be nice to keep the pipeline flexible to only run this step this if bam files are provided or if specified in another way

subworkflows/local/prepare_references.nf
fevac: I'd like to know what's best practices on that case (e.g. a file is not needed at a later step). Wasn't there an issue that files needs to be emited for cascade to work? 
In any case, I kinda like having the emit since it's makes it very clear, but happy to adapt to best practices

main.nf
fevac: agree

fevac (commented):
It looks great but this should not be part of the reference subworkflow as bam files are not reference files. See my comment below. Happy re-review when changes are made 🌟

subworkflows/local/prepare_references.nf
beatrizsavinhas: Oh I don't know actually and have no idea about the cascade issue you mentioned. But in any case, I don't think this will be relevant here anymore since the step will probably not be included in a subworkflow.

subworkflows/local/prepare_references.nf
beatrizsavinhas: Not that the parameter is required, but I meant that for there to be any cram generation there must be a bam file.

beatrizsavinhas:
> It looks great but this should not be part of the reference subworkflow as bam files are not reference files.

Yeah this is exactly what me and @kristinebilgrav discussed too! And in that case, my comments on testing the subworkflow are no longer relevant.

conf/modules/prepare_references.config
beatrizsavinhas: Looks like `oncoanalyser` uses "alignments" so we can keep it consistent. Disregard my comment!

subworkflows/local/prepare_references.nf
kristinebilgrav: As discussed, taking bam-to-cram out of prepare references, so this will be done in a different issue.

subworkflows/local/prepare_references.nf
kristinebilgrav: Have changed the logic, so it now does not have a conditional run. After discussing about channels, I think we should do that when we have figured out how to make processess optional

subworkflows/local/prepare_references.nf
kristinebilgrav: Fixed!

subworkflows/local/prepare_references.nf
kristinebilgrav: Not sure either, but after changing the logic, this is not relevant anymore I think, as the only thing we emit from the mian.nf is a multiQC file.

tests/.nftignore
kristinebilgrav: It was changing between runs for me. I am not sure what the line is, but probably a file name or date is in there

subworkflows/local/prepare_references.nf
```diff
@@ -2,7 +2,7 @@
 // Prepare reference files
 //
 
-include { UNTAR as UNTAR_VEP_CACHE                           } from '../../modules/nf-core/untar/main'
+include { UNTAR as UNTAR_VEP_CACHE } from '../../modules/nf-core/untar/main'
```
kristinebilgrav, line 5: Fixed the spaces when I removed from last review, I can add the spaces again if its annoying.

subworkflows/local/prepare_references.nf
fevac: ok, then please either open an issue or add information in the relevant existing issue

main.nf
```diff
@@ -148,8 +146,21 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
+    ch_samtools_in = ch_bam_bai.combine(ch_genome_fasta_fai)
+            .multiMap { meta_bam_bai, bam, bai, meta_fasta_fai, fasta, fai ->
+                bam_bai: tuple(meta_bam_bai, bam, bai)
+                fasta_fai: tuple(meta_fasta_fai, fasta, fai)}
+
+    SAMTOOLS_VIEW ( ch_samtools_in.bam_bai, ch_samtools_in.fasta_fai, [[], []], [[],[]], 'crai' )
+
     emit:
-    multiqc_report = ONCOREFINER.out.multiqc_report // channel: /path/to/multiqc_report.html
+    multiqc_report = ONCOREFINER.out.multiqc_report // channel: /path/to/multiqc_report.htmlk
```
fevac:
```suggestion
    multiqc_report = ONCOREFINER.out.multiqc_report // channel: /path/to/multiqc_report.html
```

fevac (approved):
Looks great! 🌟

PR #72, opened 2026-04-14
feat: Add subworkflow test for `PREPARE_REFERENCES`

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/116

### Fixed

- A subworkflow test for `PREPARE_REFERENCES`
    - Created subdirectory for the subworkflow   
    - Two tests checking the logic in the subworkflow: 
        - One where the input is an untarred directory: do not untar
        - Second where the input is a tarred file: check that its untarred by snapshotting the output
- Changed subwf input from `params.vep_cache` to `val_vep_cache`
- Removed unused channel `vep_cache_unprocessed`


subworkflows/local/prepare_references/tests/main.nf.test.snap
```diff
@@ -0,0 +1,458 @@
...
+    "Test preparing VEP cache: untarring needed": {
+        "content": [
+            [
+                "untar",
+                "untar/vep_cache",
+                "untar/vep_cache/.DS_Store",
+                "untar/vep_cache/._.DS_Store",
+                "untar/vep_cache/._LoFtool_scores.txt",
+                "untar/vep_cache/._Plugins",
+                "untar/vep_cache/._fordownload",
```
kristinebilgrav, line 25: For now checking all files, but this can be changed. We want to check that the subwf is untarring: let me know if you prefer another strategy!

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -0,0 +1,58 @@
...
+                """
+                input[0] = params.pipelines_testdata_base_path + 'reference/vep_cache_and_plugins/'
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(workflow.out).match() }
```
beatrizsavinhas:
```suggestion
                { assert snapshot(
                		workflow.out.vep_resources
                ).match() }
```

To reduce information duplication in the snapshot!

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -0,0 +1,58 @@
...
+        then {
+            // Check untarred directory and files
+            def output_directories_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)
+
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    // Check untarred directory and files
+                    output_directories_files ).match()
+                }
```
beatrizsavinhas:
```suggestion
                    output_directories_files
                ).match() }
```
Just for easier readability.

subworkflows/local/prepare_references/tests/main.nf.test.snap
beatrizsavinhas: I think this is the best way!

main.nf
beatrizsavinhas, line 1: Nice bug fixes!

beatrizsavinhas (commented):
This is a bigger question, perhaps outside of this PR as the issue currently stands, but with these tests we know that if untarring is not needed, the subworkflow simply outputs the same directory that is provided as an input. However, the outputs of the workflow in each condition are very different... Are we sure the pipeline runs correctly when an untarred directory is provided? 🤔 
Not sure about the result when using a local directory, but I'm thinking that using a GitHub directory would cause issues as it is not possible to stage a GitHub directory.

subworkflows/local/prepare_references/tests/main.nf.test.snap
```diff
@@ -0,0 +1,458 @@
...
+        "content": [
+            {
+                "0": [
+                    [
+                        "/Clinical-Genomics/test-datasets/bc94db3a30d6194f4cebcb9480bf24b0600b4d23/reference/vep_cache_and_plugins"
+                    ]
+                ],
+                "vep_resources": [
+                    [
+                        "/Clinical-Genomics/test-datasets/bc94db3a30d6194f4cebcb9480bf24b0600b4d23/reference/vep_cache_and_plugins"
```
beatrizsavinhas: I wondered if we could make this test more agnostic to the specific input path so that the test would not fail if `params.pipelines_testdata_base_path`. I thought of simply comparing the input with the output but couldn't find a feasible implementation for this. Just a thought in case you can come up with something!

kristinebilgrav:
> This is a bigger question, perhaps outside of this PR as the issue currently stands, but with these tests we know that if untarring is not needed, the subworkflow simply outputs the same directory that is provided as an input. However, the outputs of the workflow in each condition are very different... Are we sure the pipeline runs correctly when an untarred directory is provided? 🤔 Not sure about the result when using a local directory, but I'm thinking that using a GitHub directory would cause issues as it is not possible to stage a GitHub directory.

I see what you mean! Although, the output of the workflow is always just a path to a folder (at least when I have ran it). 
I tried running it with a local untarred vep_cache folder as a parameter, and it worked just fine! 
If you use the GitHub directory it will not work - thats true - but in the subworkflow test we only want to test the untarring, and for this it works!

subworkflows/local/prepare_references/tests/main.nf.test
kristinebilgrav: This somehow does not work, so changing it back to the original!

beatrizsavinhas: It works locally on my computer. Could you check again?

kristinebilgrav: Works! added again :D

beatrizsavinhas:
> > This is a bigger question, perhaps outside of this PR as the issue currently stands, but with these tests we know that if untarring is not needed, the subworkflow simply outputs the same directory that is provided as an input. However, the outputs of the workflow in each condition are very different... Are we sure the pipeline runs correctly when an untarred directory is provided? 🤔 Not sure about the result when using a local directory, but I'm thinking that using a GitHub directory would cause issues as it is not possible to stage a GitHub directory.
> 
> I see what you mean! Although, the output of the workflow is always just a path to a folder (at least when I have ran it). I tried running it with a local untarred vep_cache folder as a parameter, and it worked just fine! If you use the GitHub directory it will not work - thats true - but in the subworkflow test we only want to test the untarring, and for this it works!

Ah yes! The output is always in the format `[[<path_to_directory>]]`, it's just that the tests are checking different things for each condition. Perhaps we can make the tests more general so it's easier to see the difference between the two conditions?
I will review the PR again and add a suggestion so you can check it out!

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -0,0 +1,60 @@
...
+
+        then {
+            // Check untarred directory and files
+            def output_directories_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)
+
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    // Check untarred directory and files
+                    output_directories_files
```
beatrizsavinhas, line 62:
```suggestion
					// Check the name of the output directory
                    workflow.out.vep_resources[0][0].tokenize('/').last(),
                    // Check untarred directory and files
                    output_directories_files
```

See comment above about making both tests match.
Here collecting the directory name only avoids recording the local work directory.

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -0,0 +1,60 @@
...
+                input[0] = params.pipelines_testdata_base_path + 'reference/vep_cache_and_plugins/'
+                """
+            }
+        }
+
+        then {
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                		workflow.out.vep_resources
```
beatrizsavinhas:
```suggestion
            // Check untarred directory and files
            def output_directories_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)

            assertAll(
                { assert workflow.success },
                { assert snapshot(
                    // Check the name of the output directory
                    workflow.out.vep_resources[0][0].tokenize('/').last(),
                    // Check that no ouput files or directories are produced
                    output_directories_files
```
As mentioned in the PR discussion, perhaps it would be nice to make both tests more uniform so it's easier to see the difference between the conditions.
Additionally, if we only collect the name of the directory in the output, the test won't fail if `params.pipelines_testdata_base_path` is updated.

beatrizsavinhas (approved):

subworkflows/local/prepare_references/tests/main.nf.test
kristinebilgrav: Great - thanks!

PR #73, opened 2026-04-14
feat: Add `PROCESS_SVS` subworkflow

mathiasbio:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/120

### Added

- Subworkflow `PROCESS_SVS`.
- Subworkflow config under `conf/subworkflows/process_svs.config`.
- Bgzip to clinical and research SV filtering steps
- Subworkflow test with tumor normal data to:
	- Check output files and directories.
	- Check md5sums of stable vcf files.

### Changed

- Moved logic for processing SV VCF files to `PROCESS_SVS` subworkflow.
- Moved process settings from `conf/modules.config` to new `conf/subworkflows/process_svs.config`.
- Updated default pipelines test snapshot with new bgzipped SV vcfs
- Renamed most SV subworkflow output directories to process_svs, and modified the ignore file accordingly

### Removed

- Removed redundant if statement for `ch_sv_vcf`.


subworkflows/local/process_svs/tests/main.nf.test
```diff
@@ -0,0 +1,81 @@
...
+                """
+            }
+        }
+
+        then {
+            // All files and directories in the output directory, excluding the untar directories and files
+            def output_directories_and_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['untar**'])
+            // vcfs_with_stable_variants: Vcf files with stable variants - can be snapshot using variantsMD5, excluding the untar directories and files
+            def gz_vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'], ignore: ['untar**'])
+            def vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf'], ignore: ['untar**'])
```
beatrizsavinhas, line 74:
```suggestion
            def vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf*'], ignore: ['untar**'])
```
Would this collect all output vcf files?
Also, we should probably add this to the other subworkflow, right?

subworkflows/local/process_svs/main.nf
```diff
@@ -0,0 +1,93 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+*/
+
+//
+// MODULE: Installed directly from nf-core/modules
+//
+
+
+include { SVDB_QUERY as SVDB_QUERY_DB              } from '../../../modules/nf-core/svdb/query/main'
+include { ENSEMBLVEP_VEP as ENSEMBLVEP_SV          } from '../../../modules/nf-core/ensemblvep/vep/main'
```
beatrizsavinhas:
```suggestion
include { SVDB_QUERY                                   } from '../../../modules/nf-core/svdb/query/main'
include { ENSEMBLVEP_VEP                           } from '../../../modules/nf-core/ensemblvep/vep/main'
```
(Double check alignment.) These alias are not needed anymore, right? Also, it would be interesting to see how this affects the output directories and files in the default test snapshot. Which, by the way, you mention in the description that this PR updates but it actually doesn't haha!

workflows/oncorefiner.nf
```diff
@@ -86,85 +83,35 @@ workflow ONCOREFINER {
...
-                val_genome,
-                val_species,
-                val_vep_cache_version,
-                ch_vep_cache,
-                ch_genome_fasta,
-                ch_vep_extra_files
+        // VCF2CYTOSURE
+        ch_bam_bai = channel.empty().mix(ch_bam_bai_tumor, ch_bam_bai_normal)
+        ch_vcf2cytosure_in = ch_bam_bai.combine(
+            ch_sv_vcf.join(ch_sv_vcf_tbi, failOnMismatch: true),
```
beatrizsavinhas: This is part of the SV workflow, right?

workflows/oncorefiner.nf
```diff
@@ -34,6 +30,7 @@ include { methodsDescriptionText } from '../subworkflows/local/utils_nfcore_onco
 // LOCAL SUBWORKFLOWS
 //
 
+include { PROCESS_SVS } from '../subworkflows/local/process_svs/main.nf'
 include { PROCESS_SNVS } from '../subworkflows/local/process_snvs/main.nf'
```
beatrizsavinhas:
```suggestion
include { PROCESS_SNVS } from '../subworkflows/local/process_snvs/main.nf'
include { PROCESS_SVS    } from '../subworkflows/local/process_svs/main.nf'
```
Very minor, but it's best to follow the order of the file.

CHANGELOG.md
beatrizsavinhas, line 1: You're missing the periods haha!

beatrizsavinhas (dismissed):
Nice! I have a few comments though but the largest change needed is to include the `GENERATE_CYTOSURE_FILES` in the SV subworkflow.

subworkflows/local/process_svs/tests/main.nf.test
mathiasbio: I added this here because there were specifically VCFs created here that weren't gzipped. I don't know if we actually neeeed to check the md5 for all the VCFs or just the final output. I honestly think it's fine if we just check the final one. But here I just added for all of them because that seemed to be what we were trying to do in the SNV-subworkflow test 🤔

CHANGELOG.md
mathiasbio: NooOOoOOooo

subworkflows/local/process_svs/tests/main.nf.test
beatrizsavinhas: Yeah, I do think it's important to check all output files! So we just seemed to have missed it when working on the SNV subworkflow. That's why I suggested correcting it there too.

workflows/oncorefiner.nf
mathiasbio: Moved the cytosure steps to the SV workflow 🙏

subworkflows/local/process_svs/tests/main.nf.test
mathiasbio: Do we have some best-practices to follow for this type of thing? Like if we should check MD5 for intermediate files etc!

beatrizsavinhas: I don't know but would be interested in knowing! Please let me know if you find good resources.

kristinebilgrav (commented):
Agree with Beatriz comments, but very nice done! ⭐

subworkflows/local/process_svs/tests/main.nf.test
mathiasbio: ok! It seems like in the case of the SNV workflow all of the VCFs were compressed to begin with, so adding this extra check for uncompressed VCFs wouldn't find anything. So maybe it's probably fine if we just leave it like it is for now : )

fevac: Reading quick and sorry if I'm missing something, but all vcf we output should be gzipped. If that's not currently the case that should be fixed and not the test

mathiasbio: That makes sense @fevac 😂 I don't know, I think I just assumed that whatever way it was implemented previously was as intended. I'll keep my brain on in the future haha

subworkflows/local/process_svs/main.nf
mathiasbio: FIxed! After I figured out that I need to update the config as well 😁

subworkflows/local/process_svs/tests/main.nf.test
fevac: yeah no, everything is subject to change and some things might be far from working as intended

mathiasbio: Regarding this, I noticed that all the VCFs could be zipped by adding the extra arguments in the config, except for the one for SVDB, because that module doesn't seem to allow for this extra argument and is hardcoded to output an uncompressed VCF file. I considered for a moment to add the bgzip module to the pipeline, but it wouldn't remove the original file, just create a new compressed file 🤔

mathiasbio:
Thanks for the review guys! I think I'm ready for another round. To explain what has changed since you last looked at it:
- The generate cytosure workflow has been added to the process SV workflow as Beatriz suggested
- I then ran into some issues for the snapshot as the files are named the same in the SV and SNV workflow "subject_a_clinical.vcf" and "subject_a_clinical.vcf.gz" which after adding bgzip made them identical
- As I needed to distinguish between them somehow, and as this naming was already confusing to me, I decided to add SNV and SV to the filename to make it clearer
- This change however made it so the workflow tests which "id:SNV" and "id:SV" in the inputs created files that were named like "SV_SV_clinical.vcf.gz" and after looking at the way that this id is parsed in the general test, I saw the id would be more correctly set to "subject_a". 
- Finally, I thought it was confusing that we had different output directory names for the SV and SNV workflow for steps that were quite overlapping. Basically there was "clinical" for SV and "clinical_filtering" for the SNV, and "vep" for SNVs but "annotate_sv" for the SVs. So I tried to synchronise these. But I don't know if you agree!

subworkflows/local/process_svs/tests/main.nf.test
fellen31: Nice thinking about snapshotting everything!  I don't think you should snapshot what is put in the output directory for a subworkflow test though (this should only be done for the pipeline tests). What should be snapshot is the rather the output (emit) of the subworkflow, so I would add some to this subworkflow, even if you don't use the output further. I think this would make it easer to write the assertions. We can have a chat about if you want! :)

mathiasbio: A chat would be lovely @fellen31 ! I'll reach out on slack

beatrizsavinhas:
Nice changes @mathiasbio!
I just have a few comments before an actual review!

> * I then ran into some issues for the snapshot as the files are named the same in the SV and SNV workflow "subject_a_clinical.vcf" and "subject_a_clinical.vcf.gz" which after adding bgzip made them identical
> * As I needed to distinguish between them somehow, and as this naming was already confusing to me, I decided to add SNV and SV to the filename to make it clearer

Yes, these issues were exactly what I was thinking about when [I commented in the previous review about the process aliases](https://github.com/Clinical-Genomics/oncorefiner/pull/73#discussion_r3093645261).
This is however a side effect issue of this PR so I think we should keep the changes minimal to address it. Changing every file name requires a lot of changes to the config which will also makes it quite different than the `SNV` subworkflow config. Additionally, we have already planned to discuss these changes in more detail in [Clean up pipeline output structure and file naming](https://github.com/Clinical-Genomics/MTP-oncoflow/issues/92).

So I would argue that, for now, a quick fix of merely adding an output directory for this subworkflow (which would mean only a small block of code) would be better. And we could mention/suggest this solution in the issue above too!

> * This change however made it so the workflow tests which "id:SNV" and "id:SV" in the inputs created files that were named like "SV_SV_clinical.vcf.gz" and after looking at the way that this id is parsed in the general test, I saw the id would be more correctly set to "subject_a".

I think this change is correct! But, just for a clarification, it comes from the fact that the vcf channel meta is currently built with `id:vcf.simpleName` and the files in the test data were recently updated.
https://github.com/Clinical-Genomics/oncorefiner/blob/82b3747c301bef37ea80488f09add1c5286c8859/main.nf#L67 
Previously they were called `SV<something>.vcf.gz` and were now changed to `subject_a`. We had written the previous tests to emulate the previous conditions, meaning that this will be an issue for all tests already in place that use these channels.

Can you open an issue to document this bug and your solution so we can fix all tests in the future?

> * Finally, I thought it was confusing that we had different output directory names for the SV and SNV workflow for steps that were quite overlapping. Basically there was "clinical" for SV and "clinical_filtering" for the SNV, and "vep" for SNVs but "annotate_sv" for the SVs. So I tried to synchronise these. But I don't know if you agree!

I think this change is out of scope for this PR and we have already planned to address this issue in [Clean up pipeline output structure and file naming](https://github.com/Clinical-Genomics/MTP-oncoflow/issues/92) too. So I would suggest not adding these changes here for now.

mathiasbio:
Thanks for checking @beatrizsavinhas ! Before starting the review, then let me see if I can address these comments 😂

subworkflows/local/process_svs/main.nf
```diff
@@ -0,0 +1,118 @@
...
+        ch_sv_dbs
+            .splitCsv ( header:true )
+            .multiMap { row ->
+                vcf_dbs:  row.filename
+                in_frqs:  row.in_freq_info_key
+                in_occs:  row.in_allele_count_info_key
+                out_frqs: row.out_freq_info_key
+                out_occs: row.out_allele_count_info_key
+            }
+            .set { ch_svdb_dbs }
```
fellen31, line 54: May be out of scope of this PR, but consider using nf-schema to validate and construct `ch_sv_dbs` instead of custom csv parsing ([example schema](https://github.com/genomic-medicine-sweden/nallo/blob/5a9ad3ae07a7027bd4d271bd46699493c62fac79/assets/svdb_query_vcf_schema.json)). This allows for imposing patterns for the info keys and checking if the vcf_dbs files actually exists when starting the pipeline.

mathiasbio: I'll make an issue for it. I find it hard to know when to draw the line for the scope of a PR, I tend personally to cram in a little bit too much, but I'm learning that this is maybe not the philosophy of our team haha

mathiasbio:
Hello! @beatrizsavinhas I think I'm ready now for another round! Maybe we can discuss the suggestions @fellen31 made with only testing the emit files in the team and make a refactoring later. What do you think?

tests/default.nf.test.snap
```diff
@@ -98,13 +94,17 @@
             ],
             [
                 [
-                    "subject_a_svdbquery_vep.vcf.gz",
+                    "subject_a_clinical.vcf.gz",
```
kristinebilgrav, line 89: I think we should give this a different name so its not so confusing? Same with the research one. A suggestion is to just add SV for now and then we can refactor the whole output strategy/naming later?

mathiasbio: I agree it's impossible to know now almost which file corresponds to which workflow. If we don't do this now I would love to do this soon in a different PR. 

I actually had refactored this earlier so that each file contained SNV or SV in the name. I wouldn't be against refactoring it here, and then changing it again if we decide we want different names. I can limit it to the SV workflow, and then it wouldn't feel out of scope maybe, as you say

kristinebilgrav: Ah I see! Up to you what you prefer, but unless we do the changing of names very soon I vote adding SV is at least nice until we do that.

beatrizsavinhas: As I suggested above, I would argue for implementing a quick fix now by just adding an output directory for this subworkflow (which would mean only a small block of code whereas changing each individual file is a lot of changes). Then in the next MOB we will discuss output directories and file names!

mathiasbio: Sorry I'm not sure I understand. Probably I'm missing some terminology. So we have these output directories for instance: "research", "clinical", "annotate_sv", "svdb", which are distinguishing these files from the SNV. So they're not overwritten or anything. But without changing the filename these filenames would be the same in the general pipeline test for both workflows:

```
            [
                    "subject_a_SV_research.vcf.gz",
                    "VcfFile [chromosomes=[11, 19, 9], sampleCount=2, variantCount=6, phased=false, phasedAutodetect=false]"
                ],
                [
                    "subject_a_research.vcf.gz",
                    "VcfFile [chromosomes=[11, 22, 12, 13, 15, 16, 17, 18, 19, 1, 2, 3, 4, 5, 6, 7, 8, X, 9, 20, 21], sampleCount=2, variantCount=102, phased=false, phasedAutodetect=false]"
                ],
```

kristinebilgrav: I think its nice to be able to distinguish them like this in the snapshot. Maybe Im missing if this will make it more complex to fix later, but otherwise will be good to see which files are which until then.

beatrizsavinhas: The same output files are already tested in the subworkflow test, right?
And the subworkflow is the focus of this PR. So I don't think we need to worry much about the default test for now, since we can trust the subworkflow test. The default test will be updated when we fix all output directories and file names in the next MOB!

subworkflows/local/process_svs/tests/main.nf.test
```diff
@@ -0,0 +1,88 @@
...
```
beatrizsavinhas, line 74:
```suggestion
            def vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf*'], ignore: ['untar**'])
```
I still think all output files could be collected with this single expression.

tests/default.nf.test.snap
mathiasbio: Reverted now so that there's no changes in either the directory structure or the filenames, and I don't think that there's any conflicts with the filenames in the pipeline test (that is, they are identical in some cases such as for the research and clinical vcf, but at least they are both listed in that case). 

I guess we can live with the confusion at the moment and we can distinguish between the files in the snapshot form the pipeline test later 😂 Really itching in my fingers to clean up this confusing mess though haha

tests/default.nf.test.snap
```diff
@@ -11,11 +11,6 @@
...
                     "bcftools": 1.22
                 },
-                "ENSEMBLVEP_SV": {
-                    "ensemblvep": 115.2,
-                    "perl-math-cdf": 0.1,
-                    "tabix": 1.21
-                },
                 "ENSEMBLVEP_VEP": {
                     "ensemblvep": 115.2,
                     "perl-math-cdf": 0.1,
```
beatrizsavinhas, line 13: Interesting! So if the process names are the same, they only appear once. Makes sense!

tests/default.nf.test.snap
```diff
@@ -98,13 +94,17 @@
...
                 [
-                    "subject_a_svdbquery_vep.vcf.gz",
+                    "subject_a_clinical.vcf.gz",
                     "VcfFile [chromosomes=[], sampleCount=2, variantCount=0, phased=true, phasedAutodetect=true]"
                 ],
                 [
                     "subject_a_clinical.vcf.gz",
                     "VcfFile [chromosomes=[11, 22, 12, 13, 15, 16, 17, 18, 19, 1, 2, 3, 4, 5, 6, 7, 8, X, 9, 20, 21], sampleCount=2, variantCount=102, phased=false, phasedAutodetect=false]"
                 ],
+                [
```
beatrizsavinhas, line 96: Are we now generating more files in the pipeline? Wondering why there are these additions without anything else being removed from the snapshot. 🤔

mathiasbio:
@beatrizsavinhas @kristinebilgrav I'll happily take another review now! I think the scope is very limited now to just moving the SV file, and to avoid clashes with putting files with the same filename into the same directory I made a process_sv subdir for those processes as Beatriz suggested.

tests/default.nf.test.snap
kristinebilgrav, line 1: 👍

workflows/oncorefiner.nf
kristinebilgrav, line 1: 🚀

kristinebilgrav (approved):
Nice! Well done! 💯

workflows/oncorefiner.nf
mathiasbio: Woho! Thank you!

conf/subworkflows/process_svs.config
```diff
@@ -0,0 +1,58 @@
...
+            '--hgvs --humdiv --max_sv_size 248956422 --merged',
+            '--no_progress --numbers --per_gene --polyphen p',
+            '--protein --offline --regulatory --sift p',
+            '--symbol --tsl --uniprot --vcf'
+        ].join(' ') }
+        ext.prefix = { "${meta.id}_svdbquery_vep" }
+        publishDir = [
+            path: { "${params.outdir}/vep" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
```
beatrizsavinhas, line 44: These results will be stored outside the `/process_svs` directory. The same with the `vcf2cytosure` files!

subworkflows/local/generate_cytosure_files/tests/main.nf.test.snap
beatrizsavinhas, line 1: The changes to both these snapshots are not relevant to this PR. In the future please try to restore files to dev version in this situation.

subworkflows/local/process_svs/main.nf
```diff
@@ -0,0 +1,118 @@
...
+        ch_bam_bai_tumor      // channel: [mandatory]  [val(meta), path(bam), path(bai)]
+        ch_sv_vcf             // channel: [required]  [val(meta), path(vcf)]
+        ch_sv_vcf_tbi         // channel: [required]  [val(meta), path(vcf.tbi)]
+        ch_sv_dbs             // channel: [required]  path(svdb_dbs_csv)
+        val_genome            // value:   [required]  Genome build (e.g. GRCh38)
+        val_species           // value:   [required]  Species
+        val_vep_cache_version // value:   [required]  VEP cache
+        ch_vep_cache          // channel: [optional]  [val(meta), path(vep_cache)]
+        ch_genome_fasta       // channel: [optional]  [val(meta), path(fasta)]
+        ch_vep_extra_files    // channel: [optional]  [val(meta), path(vep_extra_files)]
```
beatrizsavinhas, line 41: Order inputs alphabetically!

subworkflows/local/process_svs/main.nf
```diff
@@ -0,0 +1,118 @@
...
+*/
+
+//
+// MODULE: Installed directly from nf-core/modules
+//
+
+include { BCFTOOLS_VIEW as BCFTOOLS_VIEW_RESEARCH   } from '../../../modules/nf-core/bcftools/view/main'
+include { BCFTOOLS_VIEW as BCFTOOLS_VIEW_CLINICAL   } from '../../../modules/nf-core/bcftools/view/main'
+include { ENSEMBLVEP_VEP                           } from '../../../modules/nf-core/ensemblvep/vep/main'
+include { SVDB_QUERY                               } from '../../../modules/nf-core/svdb/query/main'
```
beatrizsavinhas, line 14: Check the order of the imports to match the order of the code below.

beatrizsavinhas (commented):
I see that the PR has already been merged when I was out of office, but I had some comments in a pending review as I was waiting for you to finish the changes to request another review. So I thought this could still be relevant information for future PRs!

Overall, good job! It was a lot of back and forward but you got through it and made it work! 🌟

subworkflows/local/process_svs/tests/main.nf.test
mathiasbio: I see! Sorry you didn't get a chance to post them! I wasn't sure what was the proper etiquette here. 

I think I considered that, but noticed that it would get the tbi files as well! 🤔

tests/default.nf.test.snap
mathiasbio: I'm not sure if I'm understanding exactly. But at least there will be more files detected in the pipeline test as I added the bgzip to some of the SV VCF files and the regex in the pipeline test I think was only looking at the *vcf.gz.

conf/subworkflows/process_svs.config
mathiasbio: I agree! I was just trying to adhere to the scope-policy and only add the process_sv to fix the conflict with duplicate filename issue for the affected rules 😂 but maybe I overcorrected!

subworkflows/local/generate_cytosure_files/tests/main.nf.test.snap
mathiasbio: I see! I didn't know I was only supposed to update the affected snapshots but it makes sense 🤔 I'll think about this for next time

subworkflows/local/process_svs/main.nf
mathiasbio: So the processes / subworkflows should be imported in order that they're used, and input files should be ordered alphabetically? Is this according to some official guidelines from somewhere?

Is it meant that the order of the processes / subworkflows are supposed to give some type of overview over the steps maybe? I guess I can see that

beatrizsavinhas: I don't think there are specific Nextflow guidelines about this, or at least I haven't seen it.
The alphabetically order of inputs I have seen being done in several pipelines, especially when there is a large number of inputs.
The order of the imports is just a general programming recommendation.
These things just make it easier to find your way around when writing code because you can look for the placement of the input just by its name and follow the modules/subworkflows that are imported just by their order, in case you need to modify anything in the future.

subworkflows/local/process_svs/tests/main.nf.test
beatrizsavinhas: Ah alright, makes sense!

tests/default.nf.test.snap
beatrizsavinhas: Yeah, exactly. The pipeline test captured only the gzipped files and since you added that to some files that weren't zipped before, now there are more files. Mystery solved! 🕵️

PR #81, opened 2026-04-27
refactor: update publishing strategy for `process_snvs` subworkflow

kristinebilgrav:
Part of https://github.com/Clinical-Genomics/CancerBioInfo/issues/112.

### Added

- Emit block with files to publish to:
	- `PROCESS_SNVS` subworkflow.
	- `ONCOREFINER` workflow.
	- `CLINICALGENOMICS_ONCOREFINER` workflow.
	
- Output block for main workflow and corresponding publish block in `main.nf`.

- Logic for creation of conditional `ch_cadd_*` output from `ANNOTATE_CADD` in `PROCESS_SNVS` subworkflow.

### Changed

- Update `PROCESS_SNVS` subworkflow test to capture the outputs instead of published files.
- Update all relevant snapshots.

### Removed

- `publishDir` blocks from config files.


conf/subworkflows/process_snvs.config
beatrizsavinhas, line 1: @fevac We decided to do this bug fix in another PR. @mathiasbio will open it soon.

beatrizsavinhas: Fixed now.

main.nf
```diff
@@ -239,6 +250,30 @@ workflow {
...
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.cadd_annotated_tbi)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_clinical_filtered_vcf)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_clinical_filtered_tbi)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_vcfanno_vcf)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_vcfanno_tbi)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_vep_annotated_vcf)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_vep_annotated_tbi)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_vep_report)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_research_filtered_vcf)
+                    .mix(CLINICALGENOMICS_ONCOREFINER.out.snv_research_filtered_tbi)
```
beatrizsavinhas: Fix indentation and check order.

main.nf
```diff
@@ -251,6 +262,30 @@ workflow {
...
+
+    publish:
+        snv = ch_snv_publish
```
fellen31: Indentation

beatrizsavinhas: Fixed!

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -77,16 +77,24 @@ nextflow_workflow {
...
+                    workflow.out.clinical_filtered_vcf.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]},
+                    workflow.out.clinical_filtered_tbi.collect { meta, tbi -> [meta, file(tbi).name] },
+                    workflow.out.vcfanno_vcf.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]},
+                    workflow.out.vcfanno_tbi.collect { meta, tbi -> [meta, file(tbi).name] },
+                    workflow.out.vep_annotated_vcf.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]},
+                    workflow.out.vep_annotated_tbi.collect { meta, tbi -> [meta, file(tbi).name] },
+                    workflow.out.vep_report.collect { meta, _process, _tool, report -> [meta, file(report).name] },
+                    workflow.out.research_filtered_vcf.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]},
+                    workflow.out.research_filtered_tbi.collect { meta, tbi -> [meta, file(tbi).name] },
```
beatrizsavinhas: Wonder if there is a way to reduce code duplication here. 🤔

beatrizsavinhas (approved):
I think this is ready to merge! 🧹

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: Hmm, that would be nice! But dont know how, unless its possible to make some sort of loop

kristinebilgrav:
Cant approve this, but looks good to me!

fevac (approved):

subworkflows/local/process_snvs/tests/main.nf.test
fellen31: Could use a function, there are some examples in nft-utils I believe.

beatrizsavinhas: Another option is to collect all vcf files in a channel and all other files in another, then simply apply the different logics for the 2 channels.

kristinebilgrav: Hmm, that channel would only be for the tests? I think I would prefer a function if so.

beatrizsavinhas: This logic is the same we already have for the pipeline tests for example (and many other tests). It would just be a channel for the `then` block of the test.
But feel free to write a function, of course!

kristinebilgrav: Ahh, like we had before? I though you ment in the main script 😅  But that works! Should we do it together?

beatrizsavinhas: Yes, similar to what was done before but not by parsing the output directory.
I am available to work on this together!

kristinebilgrav: @fevac @mathiasbio @fellen31 What do you think about this solution?

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -77,18 +77,20 @@ nextflow_workflow {
         }
 
         then {
-            // All files and directories in the output directory, excluding the untar directories and files
-            def output_directories_and_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['untar**'])
-            // vcfs_with_stable_variants: Vcf files with stable variants - can be snapshot using variantsMD5, excluding the untar directories and files
-            def vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'], ignore: ['untar**'])
+            // All vcfs
```
beatrizsavinhas:
```suggestion
            // All output VCF files with stable variants
```

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -77,18 +77,20 @@ nextflow_workflow {
...
+            def output_vcfs = workflow.out.findAll { key, value -> key.endsWith('_vcf')}
+                .collectMany { key, value -> value}
+
+            // All tbis
+            def output_tbis = workflow.out.findAll { key, value -> key.endsWith('_tbi')}
+                .collectMany { key, value -> value}
```
beatrizsavinhas: There are other types of files that are not tbis, for example `vep_report`. I think we should have a more general variable for all output types that are not vcf files.
```suggestion
            // All other outpus that are not VCF files
            def other_outputs = workflow.out.findAll { key, value -> !key.endsWith('_tbi')}
                .collectMany { key, value -> value}
```
Not sure of the notation here for the negation but if it doesn't work we can try other options.

kristinebilgrav: In that case we would have to also have to exclude the vcfs there, which I am not sure how to do. And for now its only one file, the html, which I think is cleaner and more intuitive to keep as is. And then in the future it can be adapted if several of the same files can be collected for the same test parameters.

conf/modules.config
```diff
@@ -11,7 +11,6 @@
 */
 
 process {
-
     publishDir = [
         path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
         mode: params.publish_dir_mode,
```
kristinebilgrav: Removing this makes the subworkflow tests for `PREPARE_REFERENCES` fail, as it still uses the publishDir. So need to wait with removing this until every subworkflow is changed I guess.

fellen31: Maybe I'm not understanding the full scope, but I would be fine with not removing this in this PR, updating the snapshots, and then merging this. This part can be removed when all processes have been moved to workflow outputs in the entire pipeline.

kristinebilgrav: After we refactored the test for `PREPARE_REFERNCES` here https://github.com/Clinical-Genomics/oncorefiner/pull/123, its no longer needed to keep this!

fellen31: OK!

subworkflows/local/process_snvs/tests/main.nf.test.snap
```diff
@@ -1,68 +1,148 @@
...
-                "vep/subject_a_vep.vcf.gz_summary.html"
-            ],
             [
                 [
+                    {
+                        "id": "subject_a"
+                    },
                     "subject_a_clinical.vcf.gz",
+                    ":variantsMD5,",
                     "c7dae0b238f6b5f4bf91c0682824f474"
```
fellen31, line 11: This looks slightly weird to me : `":variantsMD5,",` but think I like the idea.

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -165,14 +167,20 @@ nextflow_workflow {
...
 
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
-                    // All directories and files in the output directory
-                    output_directories_and_files,
+                    output_vcfs.collect { meta, vcf -> [meta, file(vcf).name ]},
+                    output_tbis.collect { meta, tbi -> [meta, file(tbi).name] },
+                    workflow.out.vep_report.collect { meta, _process, _tool, report -> [meta, file(report).name] },
```
fellen31, line 183: I don't like this style of asserting, since it's not clear which file comes from which channel in the snapshots. 

```suggestion
                    workflow.out
```

fellen31 (commented):
Final comments!

subworkflows/local/process_snvs/tests/main.nf.test
beatrizsavinhas: This would put us in the same situation as before, right? See https://github.com/Clinical-Genomics/oncorefiner/pull/81#discussion_r3323193597. Would you prefer the previous setup?

subworkflows/local/process_snvs/tests/main.nf.test.snap
kristinebilgrav: Yes, did in the beginning for me as well, but its very nice in the snapshot later haha

subworkflows/local/process_snvs/tests/main.nf.test
beatrizsavinhas: I made a typo above. Line 87 should read `!key.endsWith('_vcf')`.

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -77,18 +77,20 @@ nextflow_workflow {
...
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
-                    // All directories and files in the output directory
-                    output_directories_and_files,
-                    // VCFs with stable variants
-                    vcfs_with_stable_variants.collect{ file -> [ file.getName(), path(file.toString()).vcf.variantsMD5 ] }
+                    output_vcfs.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]},
+                    output_tbis.collect { meta, tbi -> [meta, file(tbi).name] },
+                    workflow.out.vep_report.collect { meta, _process, _tool, report -> [meta, file(report).name] },
```
beatrizsavinhas, line 93:
```suggestion
                    other_outputs.collect { meta, file -> [meta, file(file).name] },
```
I think this, with my earlier suggestion https://github.com/Clinical-Genomics/oncorefiner/pull/81#discussion_r3396939019, would allows to check for any additional file. With the current implementation, if a file is added to the subworkflow that isn't either `vcf`, `tbi` or `vep_report`, it won't be included in the snapshot.

kristinebilgrav: What I like about keeping it specific to `_vcf and _tbi` is that its very clear which files are checked and what we want to check for them. Even if we in the future want to add new files and check them, we might not want to check them all in the same way?

subworkflows/local/process_snvs/tests/main.nf.test
fellen31: For the stub test you should only need to assert `workflow.out` I think.

subworkflows/local/process_snvs/tests/main.nf.test.snap
fellen31: I ment that the formatting looked slightly off. 

```
"subject_a_clinical.vcf.gz",
"variantsMD5:",
"c7dae0b238f6b5f4bf91c0682824f474"`
```
,

```
"subject_a_clinical.vcf.gz",
"variantsMD5:c7dae0b238f6b5f4bf91c0682824f474"`
```

or simply `"subject_a_clinical.vcf.gz:c7dae0b238f6b5f4bf91c0682824f474"` would have been slightly more readable to me.. But you can count this as a nit :)

subworkflows/local/process_snvs/tests/main.nf.test
fellen31: For the full test, I would lean explicitly stating the outputs but making a function for the collect...but not a hill to die on.

subworkflows/local/process_snvs/tests/main.nf.test.snap
beatrizsavinhas: Yeah, we could make it a string instead of a list!

subworkflows/local/process_snvs/tests/main.nf.test
beatrizsavinhas: Oh that is a good idea! I think I would prefer that too. 🤔

subworkflows/local/process_snvs/tests/main.nf.test
beatrizsavinhas: Perhaps a good compromise is what @fellen31 suggested above!

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: I prefer to move forward and implement this, and have opened an issue for refactoring this [here](https://github.com/Clinical-Genomics/MTP-oncoflow/issues/81): that we can work on in the future! Feel free to add the specific suggestion there

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: See comment above!

subworkflows/local/process_snvs/tests/main.nf.test.snap
kristinebilgrav: I prefer to move forward as mentioned in another comment, but this could also be refactored at a later stage!

PR #84, opened 2026-04-27
fix: vep args keep unannotated

### LINKED ISSUE #74: VEP removes SVs it cannot annotate
### Description

When running the small test SV VCF through the SV postprocessing workflow the VCF after the VEP stage is empty. After looking in the warninglog for VEP I saw this for each variant in the VCF:

```
WARNING: line 1 skipped (9 20414424 0 C C]11:118355689] 67 PASS ASMID=4...): Chromosome 9 not found in annotation sources or synonyms; chromosome 9 does not overlap any features
WARNING: line 2 skipped (9 20414999 1 C [11:118355716[C 74 PASS ASMID=2...): Chromosome 9 not found in annotation sources or synonyms; chromosome 9 does not overlap any features
WARNING: line 3 skipped (11 118355689 2 A A]9:20414424] 67 PASS ASMID=4...): Chromosome 11 not found in annotation sources or synonyms; chromosome 11 does not overlap any features
...
```

### How to reproduce

Check the final SV VCF from the current test in oncorefiner.

### Expected behaviour

All variants regardless of if they can be annotated or not should be preserved in the VCF

### Anything else?

In balsamic we're running VEP with a few extra settings such as "--allow_non_variant --dont_skip" which might be what we need to add to tell VEP not to remove these variants. http://www.ensembl.org/info/docs/tools/vep/script/vep_options.html

### System information

_No response_

mathiasbio:
closes https://github.com/Clinical-Genomics/oncorefiner/issues/74

### Added

- added extra arguments to VEP SV to avoid filtering out variants

### Changed

- renamed the write index argument to long name to match SNV workflow

### Fixed

-

### Removed

-


mathiasbio:
nf-core linting and nf-core template version is failing, but this should be fixed when we update the template

PR #87, opened 2026-04-29
feat: Update tests to hg38

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/36
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/26

### Changed

- Updated test dataset and paths to GRCh38 files
    - Results in more VCFanno files as the test dataset now contains more annotation files, and fewer vep plugin files as we will not use as many as in the previous dataset. 
- Updated filtering option `GNOMAD_popmax` to `GNOMAD_grpmax` to comply with updated files

### Fixed

- Updated `pipelines_testdata_base_path` to latest commit 
- Updated all test snapshots

### Removed

- Support for GRCh37 from default tests


subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -17,16 +17,16 @@ nextflow_workflow {
...
-                    file(params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam', checkIfExists: true),
-                    file(params.pipelines_testdata_base_path + 'testdata/subject_a.tumor.redux.bam.bai', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/subject_a.tumor.redux.bam', checkIfExists: true),
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/subject_a.tumor.redux.bam.bai', checkIfExists: true)
                 ])
                 input[1] = channel.of([
-                    [id:'SV'],
-                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
+                    [id:'subject_a'],
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.sv.vcf.gz.tbi', checkIfExists: true)
```
kristinebilgrav, line 25: This file was always supposed to be a tbi - we do not need the SNV file in this process

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -10,12 +10,14 @@ nextflow_workflow {
     test("Test preparing VEP cache: no untarring needed") {
 
         when {
+
             params {
                 outdir = "$outputDir"
             }
+            // Provided path to mock input (subworkflow wants value with path to directory) - should be improved in the future
```
kristinebilgrav, line 17: Created an issue for this: https://github.com/Clinical-Genomics/CancerBioInfo/issues/155

subworkflows/local/process_svs/tests/main.nf.test
```diff
@@ -70,8 +67,8 @@ nextflow_workflow {
             // All files and directories in the output directory, excluding the untar directories and files
             def output_directories_and_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['untar**'])
             // vcfs_with_stable_variants: Vcf files with stable variants - can be snapshot using variantsMD5, excluding the untar directories and files
-            def gz_vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'], ignore: ['untar**'])
-            def vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf'], ignore: ['untar**'])
+            def gz_vcfs_with_stable_variants = getAllFilesFromDir(params.outdir, include: ['**/*.vcf.gz'], ignore: ['untar**', 'vep/*.vcf.gz', 'process_svs/*clinical.vcf.gz'])
```
kristinebilgrav: The vep and process_svs vcf.gz files had different hd5 sums from time to time due to date in the header, therefore needed to remove them from this check. I vote we implement a chck like we have in the tumor-normal test where we check the content using vcf.summary (in a different PR?)?

conf/subworkflows/process_snvs.config
beatrizsavinhas, line 1: You've explained in Slack that this `GNOMADAF_grpmax` is the new name for the field in the hg38 files, but this is not documented in this PR. I think it is worth including this in the list in the description to explain why this code was changed!

conf/test_tumor_normal.config
```diff
@@ -14,13 +14,13 @@
 includeConfig 'test_base.config'
 
 params {
-    config_profile_name        = 'Test profile - Tumor only testdata'
-    config_profile_description = 'Minimal Tumor only test dataset to check pipeline function'
+    config_profile_name        = 'Test profile - Tumor normal testdata'
+    config_profile_description = 'Minimal Tumor normal test dataset to check pipeline function'
```
beatrizsavinhas:
```suggestion
    config_profile_name        = 'Test profile - Tumor Normal testdata'
    config_profile_description = 'Minimal Tumor Normal test dataset to check pipeline function'
```

subworkflows/local/annotate_cadd/tests/main.nf.test
```diff
@@ -53,10 +53,10 @@ nextflow_workflow {
                 """
                 input[0] = channel.of([
                     [id:'test'],
-                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
                     ])
                 input[1] = 'GRCh38'
-                input[2] = channel.fromPath(params.pipelines_testdata_base_path + 'reference/reference.fasta.fai', checkIfExists: true).map {it -> [[id:it.simpleName], it]  }.collect()
+                input[2] = channel.fromPath(params.pipelines_testdata_base_path + 'reference/GRCh38/GRCh38_masked_exclusions_alts_hlas_subset.fasta', checkIfExists: true).map {it -> [[id:it.simpleName], it]  }.collect()
```
beatrizsavinhas: Why was the previous input an index file `fai` but now replaced with a `fasta` file?

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: I think this bug fixes are out of scope for this PR (especially since it already has several other little corrections).

We should also document the bug itself and evaluate why it was not noticed before (the index file seems to be optional here).
I'd suggest opening a separate PR to fix the bugs and metas for the test for `GENERATE_CYTOSURE_FILES`.
We just need to create a branch and checkout these files there! I can do it too so let me know if you'd prefer that.

subworkflows/local/prepare_references/tests/main.nf.test
beatrizsavinhas: Very nice!
But I think the issue actually is "Add check if given `params.vep_cache` directory exists" and then we can decide on whatever solution we think is good to achieve that check.

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -27,7 +29,7 @@ nextflow_workflow {
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
-                    // Check the name of the output directory
+					// Check the name of the output directory
```
beatrizsavinhas: It seems like different characters were introduced here accidentally.

subworkflows/local/prepare_references/tests/main.nf.test.snap
```diff
@@ -10,442 +10,444 @@
...
             [
                 "untar",
                 "untar/vep_cache",
-                "untar/vep_cache/.DS_Store",
-                "untar/vep_cache/._.DS_Store",
-                "untar/vep_cache/._LoFtool_scores.txt",
                 "untar/vep_cache/._Plugins",
-                "untar/vep_cache/._fordownload",
                 "untar/vep_cache/._homo_sapiens_merged",
-                "untar/vep_cache/._pLI_values_107.txt",
```
beatrizsavinhas, line 27: Why are these files not produced anymore? Have names changed? Or is there anything else in the workflow or config that changed?

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -6,21 +6,22 @@ nextflow_workflow {
     tag "subworkflows"
     tag "process_snvs"
 
-    setup {
+
+    test("with tumor-normal GRCh38 data") {
```
beatrizsavinhas: What is the reason for moving the setup block to inside the test block?
Also check indentation.

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -31,44 +32,45 @@ nextflow_workflow {
...
                 input[4] = null
                 input[5] = channel.of([
                     [id:'SNV'],
-                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
                 ])
                 input[6] = channel.of([
                     [id:'SNV'],
-                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.sv.vcf.gz', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz.tbi', checkIfExists: true)
```
beatrizsavinhas, line 48: Again, why is the input file here of another type?

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -31,44 +32,45 @@ nextflow_workflow {
...
                 ])
                 input[7] = []
                 input[8] = channel.of([
-                    file(params.pipelines_testdata_base_path + 'reference/vcfanno_functions.lua', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'reference/GRCh38/vcfanno_functions.lua', checkIfExists: true)
                 ])
                 input[9] = channel.of([
-                    file(params.pipelines_testdata_base_path + 'reference/grch37_gnomad_-r2.1.1-.vcf.gz', checkIfExists: true),
-                    file(params.pipelines_testdata_base_path + 'reference/grch37_gnomad_-r2.1.1-.vcf.gz.tbi', checkIfExists: true)
+                    file(params.pipelines_testdata_base_path + 'reference/GRCh38/grch38_gnomad_reformatted_merged_v4.1_subset.vcf.gz', checkIfExists: true),
```
beatrizsavinhas, line 55: What are the differences in these input files? Why are there more files now? Have they been moved from `input[12]`?

subworkflows/local/process_snvs/tests/main.nf.test.snap
```diff
@@ -1,5 +1,5 @@
 {
-    "with tumor-normal data": {
+    "with tumor-normal GRCh38 data": {
```
beatrizsavinhas: Why add the `GRCh38` information here but not on the other tests? Is the idea to have another test for this subworkflow with GRCh37 test data?

subworkflows/local/process_svs/tests/main.nf.test
```diff
@@ -6,21 +6,22 @@ nextflow_workflow {
     tag "subworkflows"
     tag "process_svs"
 
-    setup {
+
+    test("with tumor-normal data") {
```
beatrizsavinhas, line 11: Same question as in the `process_snvs` test.

subworkflows/local/process_svs/tests/main.nf.test
beatrizsavinhas: In this test, we are already collecting only the variants with `.vcf.variantsMD5` in line 78. So the changes in md5sums should not be due to differences in header. Could you check if indeed there are changes in variant content for these files for each run?

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#73](https://github.com/Clinical-Genomics/oncorefiner/pull/73) Moved logic for processing SV VCF files, previously in `workflows/oncorefiner.nf`, to `PROCESS_SVS` subworkflow.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Renamed and refactored test config to `test_base.config` to include common parameters and files used for all tests.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Refactored default pipeline test and test profile to run a `tumor_normal` default test and updated snapshot.
+- [#87](https://github.com/Clinical-Genomics/oncorefiner/pull/87) Updated all testdata and file paths to GRCh38 files, updated snapshot.
```
beatrizsavinhas: This PR does not update testdata, only the file paths to the testdata!

beatrizsavinhas (commented):
Nicely done since this seems to be a lot of work! 🌟 

I have some comments and I think the most important thing to look through before merging this is investigating why the output vcf files, generated with the test data, are changing in content between each run!

Also, I think this PR is not a fix but a feature, since it's not fixing an existing bug but instead changing the behaviour of the code. I suggest renaming it to: "feat: Update tests to use hg38 testdata".

subworkflows/local/annotate_cadd/tests/main.nf.test
kristinebilgrav: Ah, good catch - its supposed to be a fasta.

subworkflows/local/generate_cytosure_files/tests/main.nf.test
kristinebilgrav: I can do it!

kristinebilgrav: Opened a PR here: https://github.com/Clinical-Genomics/oncorefiner/pull/90

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: The setup is typically related to the specific test and not general for the whole file (https://www.nf-test.com/docs/testcases/setup/). Could move it to another PR if it does not fit here.

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: Similar to above, we do not need the SV vcf in the SNV workflow. Seems like this was introduced in this PR: https://github.com/Clinical-Genomics/oncorefiner/pull/66/. Will add it to the PR above to fix.

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: These are the vcfanno resources, which have now been updated with more annotation files from CADD; SweGen and GnomAD (According to Clinical-Genomics/MTP-oncoflow#36 and Clinical-Genomics/MTP-oncoflow#26). Input[12] contains fewer files than before as we will not use the spliceAI files.

subworkflows/local/process_snvs/tests/main.nf.test.snap
kristinebilgrav: Good point. In the beginning the idea was to have two, but we moved away from that. I kept it since it makes it very clear that its GRCh38, but this is also visible other places. If it should be stated somewhere explicitly I think it should be in the pipeline tests (but dont think its needed), so I will remove it from here. Let me know if you think it should be stated somewhere at all.

subworkflows/local/process_svs/tests/main.nf.test
kristinebilgrav: See reply above.

subworkflows/local/prepare_references/tests/main.nf.test.snap
kristinebilgrav: We will not use SpliceAI files anymore, so a lot of these files are related to that. I tried to retain most of the same files, and excluding those that we would not use.

subworkflows/local/prepare_references/tests/main.nf.test.snap
```diff
@@ -10,442 +10,444 @@
...
-                "untar/vep_cache/Plugins/._MaxEntScan.pm",
                 "untar/vep_cache/Plugins/._SpliceAI.pm",
                 "untar/vep_cache/Plugins/._dbNSFP.pm",
                 "untar/vep_cache/Plugins/._pLI.pm",
-                "untar/vep_cache/Plugins/._plugin_config.txt",
                 "untar/vep_cache/Plugins/LoFtool.pm",
-                "untar/vep_cache/Plugins/MaxEntScan.pm",
                 "untar/vep_cache/Plugins/SpliceAI.pm",
                 "untar/vep_cache/Plugins/dbNSFP.pm",
                 "untar/vep_cache/Plugins/pLI.pm",
```
fevac, line 31: out of the scope, this can be removed in the future but it doesn't hurt for now. We might want to test some of those at some point

subworkflows/local/process_svs/tests/main.nf.test
fevac: It would be nice to know what's going on here. Maybe there are different number of variants in the input vcf file?

fevac (approved):
Nice! All good from my side once this is cleared out: https://github.com/Clinical-Genomics/oncorefiner/pull/87/changes#r3189298801

subworkflows/local/process_svs/tests/main.nf.test
kristinebilgrav: True! VEP is annotating with different order between every run, this is why - all the variants are there (although Mathias is working on making sure this will happen always in this PR: https://github.com/Clinical-Genomics/oncorefiner/pull/84/). I have added `--pick --pick_order mane_select,mane_plus_clinical,canonical,appris,tsl,biotype,ccds,rank,length,ensembl,refseq` to ensure that the order is the same every time. LMK if you guys think that should be its own PR or be incorporated into #84 ?

fellen31: I'm not sure how running with `--pick` would play with Scout's [internal "pick" order](https://github.com/Clinical-Genomics/scout/blob/main/scout/constants/so_terms.py). Let me know what you come up with!

subworkflows/local/process_snvs/tests/main.nf.test
beatrizsavinhas: There doesn't seem to be any explanation of that in either of the issues/PRs. I think it's important to note down and explain these decisions!

subworkflows/local/process_snvs/tests/main.nf.test.snap
beatrizsavinhas: In that case, I would suggest adding this information in the description of the test profiles!

subworkflows/local/process_snvs/tests/main.nf.test
kristinebilgrav: The files are written down in the issue https://github.com/Clinical-Genomics/MTP-oncoflow/issues/36, but I understand that its not very clear that they are for VCFanno. I added a comment in this PR about it.

subworkflows/local/process_svs/tests/main.nf.test
kristinebilgrav: Hm, I have no idea! Do not understand how this works...

kristinebilgrav: Update: made a PR here: https://github.com/Clinical-Genomics/oncorefiner/pull/92
My suggestion is to add the `--pick and --pick_order` argument to the subworkflow test only. In that way we can check that its the same for all tests, but dont mess with any order for future use of the file (ie Scout etc)

conf/subworkflows/process_snvs.config
```diff
@@ -26,8 +26,8 @@ process {
          prefix = { "${meta.id}_research" }
          args = """ --output-type z --write-index=tbi --include '(INFO/GNOMADAF_popmax <= 0.001 || INFO/GNOMADAF_popmax == ".")' """
          //args = """ --output-type z --include '(INFO/GNOMADAF_popmax <= 0.001 || INFO/GNOMADAF_popmax == ".") && (INFO/SWEGENAF <= 0.01 || INFO/SWEGENAF == ".")' """
-         //TO DO down the line: use above filtering parameters when annotation is in place
          // modules.config takes preference over test.config, and we currently dont have support for SWEGEN filtering in test profile, thus for now, we need to remove these from the filtering until this is in place.
```
kristinebilgrav: Note: this was moved from modules.config to this confi, but is still true. Which means that running `nextflow run main.nf --outdir results -profile docker,test` will fail as the test data now does not support `GNOMADAF_popmax` but `GNOMADAF_grpmax`

kristinebilgrav: Update: fixed in https://github.com/Clinical-Genomics/oncorefiner/pull/96/

tests/nextflow.config
fevac, line 1: These settings are added twice, here and in `subworkflows/local/process_snvs/tests/nextflow.config`. These settings should be the same in both and linked to the test profile, otherwise they might fall out of sync and therefore this is error prone. 

You might have already an issue for this (I'm a bit lost as didn't have time to read the whole thing), so you can disregard this if that's the case. Otherwise please open an issue to not forget even if we don't work on it at the moment.

fevac (approved):
Great! Well done! 🌟

subworkflows/local/process_snvs/tests/nextflow.config
```diff
@@ -11,19 +11,19 @@ process {
     withName:'.*PROCESS_SNVS:BCFTOOLS_VIEW_RESEARCH' {
```
fellen31, line 11: So they absolutely cannot have any effect on a normal pipeline run. 

```suggestion
    withName:'PROCESS_SNVS:BCFTOOLS_VIEW_RESEARCH' {
```

subworkflows/local/process_snvs/tests/nextflow.config
```diff
@@ -11,19 +11,19 @@ process {
...
-         args = """ --output-type z --write-index=tbi --include '(INFO/GNOMADAF_popmax <= 0.001 || INFO/GNOMADAF_popmax == ".")' """
-         //args = """ --output-type z --include '(INFO/GNOMADAF_popmax <= 0.001 || INFO/GNOMADAF_popmax == ".") && (INFO/SWEGENAF <= 0.01 || INFO/SWEGENAF == ".")' """
-         //TO DO down the line: use above filtering parameters when annotation is in place
+         args = """ --output-type z --write-index=tbi --include '(INFO/GNOMADAF_grpmax <= 0.001 || INFO/GNOMADAF_grpmax == ".")' """
+         //args = """ --output-type z --include '(INFO/GNOMADAF_grpmax <= 0.001 || INFO/GNOMADAF_grpmax == ".") && (INFO/SWEGENAF <= 0.01 || INFO/SWEGENAF == ".")' """
+         //TODO : use above filtering parameters when annotation is in place
       }
    }
 
     withName:'.*PROCESS_SNVS:BCFTOOLS_VIEW_CLINICAL' {
```
fellen31, line 20: So they absolutely cannot have any effect on a normal pipeline run. 

```suggestion
    withName:'PROCESS_SNVS:BCFTOOLS_VIEW_CLINICAL' {
```

tests/nextflow.config
fellen31: Not following 100% either, but if you are talking about `BCFTOOLS_VIEW_RESEARCH`/`BCFTOOLS_VIEW_CLINICAL` yes, I don't think these settings need to be repeated here at all. The pipeline test should (in my opinion) just use whatever configurations are set in `./conf/*`.

kristinebilgrav: As we talked about yesterday, the settings were resource and test specific and not pipeline specific. I agree that its easier to have it in the `conf/` files, but yesterday I got the impression that these should be general, and not cater to our specific needs, which I opened an issue for: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/51. And therefore we moved these settings to the test configs (and I think they need to be set in both to work?). I do agree that its error prone, especially with having to update both configs when they change... I did like the way of doing it in nallo where you have params that you call in the ext.args instead. Although that probably belongs in a different PR?

beatrizsavinhas: These settings are specific to our test data, so they shouldn't necessarily be the default for all runs, i.e. be included in `./conf/*`.
We thought of having them applied only for tests or test runs (with `-profile test`). The thing is that we have 2 tests that are dependant on these tools - the subworkflow test and the pipeline test. The same test data is used so, in this situation, we had to add the same setting for both test configs.
These changes were implemented here https://github.com/Clinical-Genomics/oncorefiner/pull/96 and the current PR only updates them to applied to the new test data.

Do you think these settings should be the default, in the `./conf/*`? Otherwise, how else could we have them applied to only tests? 🤔 

This issue is especially relevant for vep settings - see https://github.com/Clinical-Genomics/oncorefiner/pull/95 - because only in the test we require the annotation order to be the same, so we can ensure a stable md5sum.

subworkflows/local/process_snvs/tests/nextflow.config
beatrizsavinhas: Agreed! But are the subworkflow test configs ever loaded in a pipeline run?! 🤔  
In any case I also think these changes could be part of this issue https://github.com/Clinical-Genomics/CancerBioInfo/issues/144 so I would argue it is OK that they are not addressed in this PR to avoid scope creep.

tests/nextflow.config
fellen31: Wrote a suggestion here: https://github.com/Clinical-Genomics/oncorefiner/pull/96#issuecomment-4398749693

subworkflows/local/process_snvs/tests/nextflow.config
beatrizsavinhas: See [comment above](https://github.com/Clinical-Genomics/oncorefiner/pull/87/changes/BASE..aee5fbe8b13bff05b1b53e0b2f550958082e5adc#r3207011487).

subworkflows/local/process_snvs/tests/nextflow.config
fellen31: No you are right, they should not. It's the opposite that could happen, subworkflow configs from `/conf/` being applied to the subworkflow tests. Absolutely! I missed that you already had an issue for it.

PR #89, opened 2026-05-05
feat: Add markdown script for CNV report generation

fevac:
This script is part of https://github.com/Clinical-Genomics/MTP-oncoflow/issues/34

### Added

- Markdown script for CNV report generation, that takes two CNV tables (at gene and segment level) and creates a interactive report where those tables are searchable and sortable.

### Notes

- This PR displays tables as-is. The report will need to be optimised once we know which values are relevant to display after assessing the verification cases. 

- Standalone script usage:
```
Rscript -e "rmarkdown::render('cnv_report.Rmd', params = list(cnv_gene = 'path/to/cnv_gene.tsv', cnv_segment = 'path/to/cnv_segment.tsv'), output_file = 'cnv_report.html')"
```


assets/cnv_report.Rmd
fevac, line 1: Note: Changelog will be updated as part of parent issue

assets/cnv_report.Rmd
```diff
@@ -0,0 +1,91 @@
...
+<summary><strong>Input data (click to expand)</strong></summary>
+
+- **Gene-level CNV file:** `r params$cnv_gene`
+- **Segment-level CNV file:** `r params$cnv_segment`
+
+</details>
+
+
+<details>
+<summary><strong>How to use the yables (click to expand)</strong></summary>
```
fellen31:
```suggestion
<summary><strong>How to use the tables (click to expand)</strong></summary>
```

assets/cnv_report.Rmd
fellen31, line 1: If this script is to be used within a module, I would put this script within the module: https://docs.seqera.io/nextflow/modules/developing-modules#resources 😊

assets/cnv_report.Rmd
```diff
@@ -0,0 +1,91 @@
...
+    highlight: tango
+    code_folding: hide
+    toc: false
+params:
+  cnv_gene: ""
+  cnv_segment: ""
+---
+
+```{r setup, include=FALSE}
+library(readr)
```
fellen31, line 16: 💯

fellen31 (commented):
Cool!

assets/cnv_report.Rmd
```diff
@@ -0,0 +1,91 @@
...
+
+The tables below allow sorting and filtering.
+
+**Sorting** - Click any column header to sort ascending or descending.
+
+**Filtering** - Use the boxes at the top of each column to filter results. The tables support **regular expression (regex)** search:
+
+  - Exact match (recommended for genes): `^TP53$` → matches only TP53
+  - Starts with: `^BRCA`
+  - Ends with: `TP53$`
```
mathiasbio, line 46: 🌟

kristinebilgrav (approved):
Very nice!

assets/cnv_report.Rmd
beatrizsavinhas: I'm curious: How would this work if we were to use a shared module, like [nf-co.re/modules/rmarkdownnotebook](https://nf-co.re/modules/rmarkdownnotebook)? 🤔 
Do you have any idea @fellen31 ?

beatrizsavinhas: Even though the CI checks might fail, you should still be able to merge to your feature branch since it's not a protected branch! 😉

mathiasbio:
A small thing I noticed when looking at the example report was that the Gene column search was a little buggy. The filtering seemed to work but the searchbar didn't display the genename you search for. But maybe that can be addressed later!

assets/cnv_report.Rmd
fellen31: Not sure...old way of doing things was to put all scripts in a shared `./bin/` folder.

beatrizsavinhas: Ah it seems that the module itself receives the path to the file as an input, so I guess it doesn't really matter where the file is located.

But otherwise, putting it in a shared `./bin/` directory would probably be the way to go since nextflow will automatically detect it there, as far as I understand!

fellen31: Well, it's only if the script is to be used within the `script:` block (and not as an input file). If it is actually an input, then I guess `assets` is fine.

PR #90, opened 2026-05-05
fix: fix subworkflow test input and meta

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/156

### Fixed

In `GENERATE_CYTOSURE_FILES`
- Updated input[1] to be tbi of the input SV VCF (from SNV vcf)
- Updated meta to be `subject_a`
- Updated snapshot

In `PROCESS_SNVs`
- Updated input[6] to be tbi of the input SNV VCF (from SV vcf)
- Updated meta to be `subject_a`
- Updated snapshot


fevac (approved):
Good catch! Out of the scope but just a thought given that the content of meta doesn't seem to be checked in these tests, we might want to give different values that are not linked to the file name. Regardless, this will probably not be an issue once meta is updated

fellen31:
> Good catch! Out of the scope but just a thought given that the content of meta doesn't seem to be checked in these tests, we might want to give different values that are not linked to the file name. Regardless, this will probably not be an issue once meta is updated

💯 the meta of the subworkflow tests should/could be whatever and doesn't have to be linked to the filename. I would encourage scrapping snapshotting the outputdir in the subworkflow tests, and instead snapshot the output of the emit. And then, it might be easier just to include the meta in the snapshot as well.

beatrizsavinhas:
> > Good catch! Out of the scope but just a thought given that the content of meta doesn't seem to be checked in these tests, we might want to give different values that are not linked to the file name. Regardless, this will probably not be an issue once meta is updated
> 
> 💯 the meta of the subworkflow tests should/could be whatever and doesn't have to be linked to the filename. I would encourage scrapping snapshotting the outputdir in the subworkflow tests, and instead snapshot the output of the emit. And then, it might be easier just to include the meta in the snapshot as well.

For sure! We will come back and update the tests once we are done with the refactoring of the publishing strategy!

beatrizsavinhas (approved):
Looks good to me!

subworkflows/local/process_snvs/tests/main.nf.test
```diff
@@ -38,12 +38,12 @@ nextflow_workflow {
...
                 input[5] = channel.of([
-                    [id:'SNV'],
+                    [id:'subject_a'],
                     file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz', checkIfExists: true)
                 ])
                 input[6] = channel.of([
-                    [id:'SNV'],
-                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.sv.vcf.gz', checkIfExists: true)
+                    [id:'subject_a'],
+                    file(params.pipelines_testdata_base_path + 'testdata/tumor_normal/subject_a.tumor.purple.somatic.vcf.gz.tbi', checkIfExists: true)
```
beatrizsavinhas, line 46: This does not only change the input from the vcf to the corresponding tbi file, but also changes which file is used! There is no information or explanation of this in the PR or issue description. Please document.

kristinebilgrav: Do you mean that the subworkflow was using the SV vcf before? Because it was using input[5] as the input file, and then this was taken as the tbi. I tried to update the comment in the PR. LMK if I misunderstood something!

beatrizsavinhas: So what was actually happening was that on top of incorrectly using a vcf instead of an index file, the test was also using a completely different vcf file that didn't match the input! 😆

The fix was to correct the input to be the matching index file for the vcf in `input[5]`.

beatrizsavinhas:
> ## PR checklist
> * [ ]  Fill in description of the PR and link to any relevant issues.
> * [ ]  If you've fixed a bug, added a subworkflow or any other code that should be tested, add tests!
> * [ ]  If you've added a new tool, update citations according to the instructions under [Adding Citations](https://github.com/Clinical-Genomics/oncorefiner/tree/dev?tab=contributing-ov-file#adding-citations).
> * [ ]  Usage Documentation in `docs/usage.md` is updated.
> * [ ]  Output Documentation in `docs/output.md` is updated.
> * [ ]  `README.md` is updated (including authors/contributors).
> 
> See [Contributing Guidelines](https://github.com/Clinical-Genomics/oncorefiner/tree/dev?tab=contributing-ov-file#readme) for more details on how to contribute according to pipeline conventions.

Make sure to also look at the checklist!

PR #91, opened 2026-05-05
feat: add cnv report

fevac:
Closes: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/34

### Added

- CNV report

### Note

There is a https://nf-co.re/modules/rmarkdownnotebook module

However it was unsure about using it or not as: 
- the environment lacks R packages that are needed by the CNV report
- the main.nf add also more logic and complexity and I didn't have the time to check it. I'll check further whenever I have time

### DIFF
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -16,6 +16,8 @@ Define where the pipeline should find input data and save output data.
 | `bai_tumor` | Path to BAM index file for the tumor sample. | `string` |  |  |  |
 | `bam_normal` | Path to alignment BAM file for the normal sample. | `string` |  |  |  |
 | `bai_normal` | Path to BAM index file for the normal sample. | `string` |  |  |  |
+| `cnv_gene` | Path to a file containing gene-level CNV calls. | `string` |  |  |  |
+| `cnv_segment` | Path to a file containing segment-level CNV calls. | `string` |  |  |  |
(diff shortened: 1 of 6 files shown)


kristinebilgrav:
Could we ask them to update the environment with more packages?

beatrizsavinhas:
Thoughts about using the [nf-co.re/modules/rmarkdownnotebook](https://nf-co.re/modules/rmarkdownnotebook) module:

* There is no topic versions support, which your module has. Perhaps we could suggest the changes to nf-core?
* It seems like their code already accounts for known rmarkdown bugs and other possible issues with rendering such files. This very likely makes their code more robust. If we were to run into issues in the future, this extra logic already in place could save us troubleshooting time. Plus, if we use the nf-core module and run into related issues, we can always ask the community for help (which is harder to do when using a local module).
* Also, [they are already accounting for the probable possibility of overriding containers](https://github.com/nf-core/modules/blob/572d8b2e40a6d36e2c67bd097ab4801a43adfc4e/modules/nf-core/rmarkdownnotebook/main.nf#L5-L7) so this is likely to not be an issue with future updates either.

With this in mind, and if it indeed works well to provide a custom container through the config file, perhaps it's worth using the community module!

fevac:
Thanks for the comments! Yes! I'll give it a try to the nfcore module and see if it works, but will test in different PR. Thanks for the early feedback

fevac:
Closed in favour of https://github.com/Clinical-Genomics/oncorefiner/pull/126

PR #92, opened 2026-05-06
fix: enforce order of vep annotation

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/48

### Changed

- Added nextflow.config with VEP `--pick --pick_order` arguments to `PROCESS_SVs` subworkflow test in order to enforce same order every time (for compatibility with `variantsMD5`)


subworkflows/local/process_svs/tests/nextflow.config
```diff
@@ -0,0 +1,17 @@
+process {
+    withName: 'PROCESS_SVS:ENSEMBLVEP_VEP' {
```
beatrizsavinhas:
```suggestion
    withName: 'ENSEMBLVEP_VEP' {
```
For the test config only, this would be sufficient! Plus it makes it more versatile in case the subworkflow changes names in the future! 🙂

beatrizsavinhas (approved):
Great idea! 🌟 

This solution looks good to me but watch out that it does not fulfil the acceptance criteria of the issue (I think that should probably be generalised to allow for a different solution like this one!).

tests/nextflow.config
```diff
@@ -33,4 +33,20 @@ process {
         memory: '15.GB',
         time: '1.h'
     ]
+
+    withName: 'ENSEMBLVEP_VEP' {
```
beatrizsavinhas, line 37: Watch out that in the pipeline, `ENSEMBLVEP_VEP` is called in several different places, so with this name regex these settings are applied for all these processes!

kristinebilgrav: Good catch! Will implement!

PR #93, opened 2026-05-07
fix: Update publishing strategy for `PROCESS_SVS`

beatrizsavinhas:
### Added

- Emit block with files to publish to:
	- `GENERATE_CYTOSURE_FILES` subworkflow.
	- `PROCESS_SVS` subworkflow.
	- `ONCOREFINER` workflow.

### Changed

-

### Fixed

-

### Removed

-

### DIFF
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -12,12 +12,6 @@
 
 process {
 
-    publishDir = [
-        path: { "${params.outdir}/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
-        mode: params.publish_dir_mode,
-        saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
-    ]
-
     withName: 'MULTIQC' {
         ext.args   = { params.multiqc_title ? "--title \"$params.multiqc_title\"" : '' }
         publishDir = [
@@ -26,8 +20,6 @@ process {
             saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
         ]
     }
-
-
 
     withName: 'SAMTOOLS_VIEW' {
         ext.prefix = { "${input.simpleName}_${meta.type}" }
--- a/conf/subworkflows/process_snvs.config
+++ b/conf/subworkflows/process_snvs.config
@@ -24,7 +24,7 @@ process {
     withName:'.*PROCESS_SNVS:BCFTOOLS_VIEW_RESEARCH' {
       ext {
          prefix = { "${meta.id}_research" }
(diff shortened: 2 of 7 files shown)


subworkflows/local/process_svs/tests/main.nf.test.snap
```diff
@@ -2,72 +2,287 @@
...
+            ],
+            [
                 [
-                    "subject_a_svdbquery_vep.vcf.gz",
-                    "f92bc125c747cdbce5e5c6c2ff8b023a"
+                    {
+                        "id": "subject_a",
+                        "case_id": "test_tumor_normal_case"
+                    },
+                    "null.cgh"
```
sofiademmou, line 56: Should this be `null` ? Maybe `file(cgh).name` doesn't work here ? :)

tests/optional_inputs_stub.nf.test.snap
```diff
@@ -58,14 +58,6 @@
                 "multiqc/multiqc_report.html",
                 "pipeline_info",
                 "pipeline_info/oncorefiner_software_mqc_versions.yml",
-                "process_svs",
-                "process_svs/test_tumor_normal_case.tsv",
```
sofiademmou, line 62: You don't output the tsv anymore ? (sorry if this is obvious, i just want to make sure since i am not super familiar with oncorefiner)

subworkflows/local/process_svs/tests/main.nf.test.snap
kristinebilgrav: Ideally it should not hehe, but it was previously too. I agree that it should be fixed though, so I will open another PR to fix this, since I think its to do with the meta in the subworkflow test. Thanks!

tests/optional_inputs_stub.nf.test.snap
kristinebilgrav: Good catch! I think its because we now only snapshot the files that are actually published in emit, and this is an intermediate file which will not be published. :)

subworkflows/local/process_svs/tests/main.nf.test.snap
sofiademmou: Perfect !

tests/optional_inputs_stub.nf.test.snap
sofiademmou: Okay, I see ! Perfect then :) Thank you for your answer !

subworkflows/local/process_svs/tests/main.nf.test.snap
kristinebilgrav: Linking the issue here: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/107

PR #96, opened 2026-05-07
fix: Move custom test settings for `PROCESS_SNVS:BCFTOOLS_VIEW_*` to test configs

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/52.

### Fixed

- Moved custom `ext` settings for `PROCESS_SNVS:BCFTOOLS_VIEW_*` to test configs (subworkflow and pipeline).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#72](https://github.com/Clinical-Genomics/oncorefiner/pull/72) Added subworkflow test to `PREPARE_REFERENCES`
 - [#82](https://github.com/Clinical-Genomics/oncorefiner/pull/82) Add index files for SNV clinical and research filtered vcfs
 - [#90](https://github.com/Clinical-Genomics/oncorefiner/pull/90) Fixed bug in `GENERATE_CYTOSURE` and `PROCESS_SNVs`: updated test meta to be `subject_a` and input to contain tbi.
+- [#96](https://github.com/Clinical-Genomics/oncorefiner/pull/96) Moved custom settings for `PROCESS_SNVS:BCFTOOLS_VIEW_*` to test config.
 
 ### `Dependencies`
 
--- a/conf/subworkflows/process_snvs.config
+++ b/conf/subworkflows/process_snvs.config
@@ -21,6 +21,14 @@ process {
         ]
     }
 
+    withName:'.*PROCESS_SNVS:BCFTOOLS_VIEW_RESEARCH' {
+      publishDir = [
+            path: { "${params.outdir}/research_filtering" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
+        ]
+   }
+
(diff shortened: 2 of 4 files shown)


kristinebilgrav (approved):
Amazing! 🌟

fellen31:
Sorry, but I think this approach will be quite error prone :/ Especially as things get more complicated, like:

```
withName: '.*:PROCESS_SVS:ENSEMBLVEP_VEP' {
        ext.args = { [
            '--dir_cache cache',
            '--dir_plugins cache/Plugins',
            '--plugin pLI,pLI_values.txt',
            '--appris --biotype --buffer_size 100 --canonical --cache --ccds',
            '--compress_output bgzip --distance 5000 --domains',
            '--exclude_predicted --force_overwrite --format vcf',
            '--hgvs --humdiv --max_sv_size 248956422 --merged',
            '--no_progress --numbers --per_gene --polyphen p',
            '--protein --offline --regulatory --sift p',
            '--symbol --tsl --uniprot --vcf',
            '--pick --pick_order mane_select,mane_plus_clinical,canonical,appris,tsl,biotype,ccds,rank,length,ensembl,refseq'
        ].join(' ') }
```

Which is hard to keep in sync with all arguments, as I think @fevac was taking about in [this comment ](https://github.com/Clinical-Genomics/oncorefiner/pull/87#discussion_r3202634203).

There might be specific changes that you need to do to the test profile to make your pipeline run with the test data (e.g. tools that by default are too memory intensive to run on GitHub runners, tools that fail if certain regions are missing). One way to solve it is to add extra arguments just for the test profile(s) to the process configs in `./conf/*` that are empty by default, but can be modified by the test config, see for example: https://github.com/genomic-medicine-sweden/nallo/blob/75b3a637dd09c03861352ef263a89e5e7377aa2c/conf/modules/genome_assembly.config#L27-L35 and https://github.com/genomic-medicine-sweden/nallo/blob/75b3a637dd09c03861352ef263a89e5e7377aa2c/tests/nextflow.config#L66-L72.

Happy to hear if you find a better solution!

beatrizsavinhas:
> One way to solve it is to add extra arguments just for the test profile(s) to the process configs in `./conf/*` that are empty by default, but can be modified by the test config, see for example: [genomic-medicine-sweden/nallo@`75b3a63`/conf/modules/genome_assembly.config#L27-L35](https://github.com/genomic-medicine-sweden/nallo/blob/75b3a637dd09c03861352ef263a89e5e7377aa2c/conf/modules/genome_assembly.config#L27-L35) and [genomic-medicine-sweden/nallo@`75b3a63`/tests/nextflow.config#L66-L72](https://github.com/genomic-medicine-sweden/nallo/blob/75b3a637dd09c03861352ef263a89e5e7377aa2c/tests/nextflow.config#L66-L72).
> 
> Happy to hear if you find a better solution!

Ahhh I see! This could be a way to go about it! And it relates to this other issue as well - https://github.com/Clinical-Genomics/CancerBioInfo/issues/108. I agree that it would be a better solution than what we currently have, avoiding the maintenance overhang and risk of diverging settings. Would like to hear the thoughts of other's in the team about this!

Thank you @fellen31!!!

PR #97, opened 2026-05-12
feat: Add `VCF_ANNOTATE_SCORE_GENMOD` to `PROCESS_SNVS`

beatrizsavinhas:
### Added

- `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow and included modules:
	- `genmod/annotate`
	- `genmod/compound` 
	- `genmod/models`
	- `genmod/score`
- `conf/subworkflows/vcf_annotate_score_genmod.config` with the following settings:
   - `GENMOD_SCORE`
      - `--family_id ${meta.id}` - so that the score annotation in the output file shows `RankScore=<meta.id>:<score>`.
      - `--rank_results` - that provides an additional `RankResult` field in the output vcf file which has a score breakdown by category.
    - `BCFTOOLS_VIEW`
       - `--output-type z` - to produce a zipped output vcf file.
       - `--write-index tbi` - to produces an index `tbi` file.

### Changed

- `nf-test.config` to also ignore tests from `subworkflows/genomic-medicine-sweden`.
- Updated `bcftools/view` module included in `vcf_annotate_score_genmod` subworkflow.

------

:x: Pipeline tests addressed in https://github.com/Clinical-Genomics/oncorefiner/pull/106 (dependent on https://github.com/Clinical-Genomics/test-datasets/pull/5). See [Development overview](https://github.com/Clinical-Genomics/MTP-oncoflow/issues/31#issuecomment-4438519373) for reference.


subworkflows/local/process_snvs/main.nf
beatrizsavinhas, line 1: This is the critical decision of this PR. Is it correct that the input file for scoring is the VEP annotated file? And then that the GENMOD ranked file is the input for clinical filtering?

fellen31: > This is the critical decision of this PR. Is it correct that the input file for scoring is the VEP annotated file?

I would say this part should probably be true, there might be things one would want to score that is annotated by VEP.

conf/subworkflows/vcf_annotate_score_genmod.config
```diff
@@ -0,0 +1,22 @@
...
+        ext.prefix = File name prefix for output files.
+----------------------------------------------------------------------------------------
+*/
+
+process {
+
+    withName:'.*VCF_ANNOTATE_SCORE_GENMOD:GENMOD_SCORE' {
+         ext.args = { [
+            "--family_id ${meta.id}",
+            "--rank_results"
```
beatrizsavinhas, line 18: This flag provides an additional `RankResult` field in the output vcf file which has a score breakdown by category.

conf/subworkflows/vcf_annotate_score_genmod.config
```diff
@@ -0,0 +1,22 @@
...
+        ext.args   = Additional arguments appended to command in module.
+        ext.args2  = Second set of arguments appended to command in module (multi-tool modules).
+        ext.args3  = Third set of arguments appended to command in module (multi-tool modules).
+        ext.prefix = File name prefix for output files.
+----------------------------------------------------------------------------------------
+*/
+
+process {
+
+    withName:'.*VCF_ANNOTATE_SCORE_GENMOD:GENMOD_SCORE' {
```
fellen31:
```suggestion
    withName:'.*:VCF_ANNOTATE_SCORE_GENMOD:GENMOD_SCORE' {
```

main.nf
```diff
@@ -38,6 +38,7 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     val_bai_tumor                   // string:  [optional]  path to BAI file for the tumor sample
     val_cadd_prescored_indels       // string:  [optional]  path to CADD prescored indels file
     val_cadd_resources              // string:  [optional]  path to CADD resources directory
+    val_genmod_score_config         // string:  [optional]  path to Genmod config file
```
fellen31: Perhaps
```suggestion
    val_genmod_score_config         // string:  [optional]  path to Genmod score config file
```

main.nf
```diff
@@ -139,6 +140,9 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     ch_sv_dbs            = val_svdb_query_dbs    ? channel.fromPath(val_svdb_query_dbs)
                                                  : channel.empty()
 
+    // Input for genmod_score
+    ch_genmod_score_config = val_genmod_score_config ? channel.fromPath(val_genmod_score_config).map { it -> [[id:'genmod_score_config'], it] }.collect()
```
fellen31: Could use a more descriptive name than "it". And Perhaps use the filename as id, if you wish. 

```suggestion
    ch_genmod_score_config = val_genmod_score_config ? channel.fromPath(val_genmod_score_config).map { file_path -> [[id:file_path.simpleName], file_path] }.collect()
```

fellen31 (commented):
🤜 🤛

main.nf
beatrizsavinhas: Well, even though your suggestion would be better, both of these will not work because the subworkflow joins the `ch_vcf` with `ch_genmod_score_config` expecting them to have the same meta. 😅 

I was going to change this soon or possibly wait for when we refactor the input file strategy. But I'll try now to write something that will actually work and get back to you!

fellen31: I forgot about that. So the implementation in this PR would not work then, right? Maybe it is a tricky requirement for the subworkflow 😅 

Suggest to combine the `ch_vcf` and `ch_genmod_score_config` channels + multiMap to give the `ch_vcf` meta to both inputs then :)

main.nf
```diff
@@ -139,6 +140,12 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     ch_sv_dbs            = val_svdb_query_dbs    ? channel.fromPath(val_svdb_query_dbs)
                                                  : channel.empty()
 
+    // Input for genmod_score
+    ch_genmod_score_config = val_genmod_score_config && val_snv_vcf ?
+                                                     channel.fromPath(val_genmod_score_config)
+                                                     .combine(ch_snv_vcf)
+                                                     .map { genmod_score_config, meta, _snv_vcf -> [meta, genmod_score_config] }
+                                                     : channel.empty()
```
beatrizsavinhas: This is absolutely not great, but the subworkflow joins the `ch_vcf` with `ch_genmod_score_config` expecting them to have the same meta so this logic is currently necessary. In the future, this will be easily fixed by a better input strategy as per https://github.com/Clinical-Genomics/MTP-oncoflow/issues/42.

beatrizsavinhas: What do you think of this @fellen31? 😅

main.nf
beatrizsavinhas: See https://github.com/Clinical-Genomics/oncorefiner/pull/97/changes#r3279630509! :)

conf/subworkflows/vcf_annotate_score_genmod.config
beatrizsavinhas: Added this information to the description of the PR.

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: Yes, that was my thought as well. And then the score is just added as additional annotation to vcf file so I thought it made sense to carry that over to the clinical filtering.
But just wanted to highlight this decision so the rest of the team is aware!

kristinebilgrav: I agree.

subworkflows/local/process_snvs/main.nf
```diff
@@ -14,6 +14,12 @@ include { BCFTOOLS_VIEW as BCFTOOLS_VIEW_RESEARCH } from '../../../modules/nf-co
 include { BCFTOOLS_VIEW as BCFTOOLS_VIEW_CLINICAL } from '../../../modules/nf-core/bcftools/view/main'
 include { ANNOTATE_CADD                           } from '../../../subworkflows/local/annotate_cadd'
 
+//
+// SUBWORKFLOW: Installed directly from genomic-medicine-sweden/subworkflows
```
kristinebilgrav, line 18: Perhaps make this comment more general in case we add other subworkflows in the future?

subworkflows/local/process_snvs/main.nf
```diff
@@ -99,13 +106,17 @@ workflow PROCESS_SNVS {
             ch_vep_extra_files
         )
 
+        VCF_ANNOTATE_SCORE_GENMOD (
+            ENSEMBLVEP_VEP.out.vcf,
+            channel.empty(),
+            channel.empty(),
+            ch_genmod_score_config,
+            true
+        )
```
kristinebilgrav, line 123: Nice!

main.nf
kristinebilgrav: I am a little confused, according what I understand from the subworkflow, this channel is supposed to hold only an ini file. Why are you joining it with a vcf here? If it requires the same meta, would it not be better to do this in the process_snvs, especially since you want to use the VEP output and not this one?

fellen31: Think Kristine has a point, that this might be safer to do closer to the actual subworkflow, in case the meta of the SNV VCF would change inside oncorefiner.

beatrizsavinhas: Initially I had thought to put this logic here since this is where channels are initialised. But you do have a point in that the correct thing to do is make sure the meta matches the input vcf channel that is actually provided to `vcf_annotate_score_genmod`! So I'll move this logic to the `process_snvs` subworkflow instead :arrows_counterclockwise:

There should also be logic to make only validate the `val_genmod_score_config` if `val_snv_vcf` but that could be handled in https://github.com/Clinical-Genomics/MTP-oncoflow/issues/64 instead.

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: Thanks! 🌟

subworkflows/local/process_snvs/main.nf
```diff
@@ -99,13 +106,25 @@ workflow PROCESS_SNVS {
             ch_vep_extra_files
         )
 
+        // Rank and add score annotation with genmod_score
+        ch_annotate_score_genmod_in = ENSEMBLVEP_VEP.out.vcf
+            .combine(ch_genmod_score_config)
+            .multiMap { meta_vcf, vcf, _meta_genmod_score_config, genmod_score_config ->
+                vcf: tuple(meta_vcf, vcf)
+                score_config: tuple(meta_vcf, genmod_score_config)
+            }
```
beatrizsavinhas, line 115: In sequence of https://github.com/Clinical-Genomics/oncorefiner/pull/97#discussion_r3279630509, @kristinebilgrav and @fellen31 see if you like this solution!

main.nf
beatrizsavinhas: See https://github.com/Clinical-Genomics/oncorefiner/pull/97/changes#r3281516265.

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: I tried to keep the same format as the previous comment in lines 7-9. Perhaps it's useful to have a separation for each source of module/subworkflow? 🤷‍♀️

subworkflows/local/process_snvs/main.nf
kristinebilgrav: Yes, great!

main.nf
```diff
@@ -139,6 +140,9 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     ch_sv_dbs            = val_svdb_query_dbs    ? channel.fromPath(val_svdb_query_dbs)
                                                  : channel.empty()
 
+    // Input for genmod_score
+    ch_genmod_score_config = val_genmod_score_config ? channel.fromPath(val_genmod_score_config).map { it -> [[id:it.simpleName], it] }.collect()
```
fevac, line 144: I guess the meta id will be changed once the new strategy is defined

subworkflows/local/process_snvs/main.nf
fevac: yes! agree!

fevac (approved):
Nice! 🌟

kristinebilgrav (approved):
🚀 ⭐

main.nf
beatrizsavinhas: Maybe, but probably not because this is just a reference file, not a sample file that would go in a samplesheet. 🤔

PR #100, opened 2026-05-13
fix: Template update for nf-core/tools v4.0.2

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/154

### Changed

- Update template for nf-core/tools v4.0.2.
- Updated `multiqc` module (included in new template).
- Changed minimum required nextflow version to 25.10.4 (included in new template).
- Added `CODE_OF_CONDUCT.md` to lint ignore list. The file has been removed since the organisation is not nf-core (included in new template) but nf-core linting still expects the file to exist.
- Added `.github/ISSUE_TEMPLATE/config.yml` to lint ignore list to comply with updated linting requirements.
- Added the following parameters as input to `CLINICALGENOMICS_ONCOREFINER` as they are now required by `ONCOREFINER`:  
	```
	multiqc_config
    multiqc_logo
    multiqc_methods_description
    outdir
	```


workflows/oncorefiner.nf
```diff
@@ -3,30 +3,13 @@
...
 include { methodsDescriptionText } from '../subworkflows/local/utils_nfcore_oncorefiner_pipeline'
-
-//
-// LOCAL SUBWORKFLOWS
-//
-
-include { PROCESS_SNVS } from '../subworkflows/local/process_snvs/main.nf'
-include { PROCESS_SVS } from '../subworkflows/local/process_svs/main.nf'
+include { PROCESS_SNVS           } from '../subworkflows/local/process_snvs/main.nf'
+include { PROCESS_SVS            } from '../subworkflows/local/process_svs/main.nf'
```
beatrizsavinhas, line 12: The updated template seems to not contain the comment so I removed the "LOCAL SUBWORKFLOWS" comment too.

.pre-commit-config.yaml
```diff
@@ -17,15 +17,21 @@ repos:
...
               .*\.snap$
           )$
       - id: end-of-file-fixer
         exclude: |
           (?x)^(
               .*ro-crate-metadata.json$|
-              modules/nf-core/.*|
-              subworkflows/nf-core/.*|
+              modules/(?!local/).*|
+              subworkflows/(?!local/).*|
```
beatrizsavinhas, line 29: New settings account for all shared modules/subworkflows (not just nf-core)!

.nf-core.yml
```diff
@@ -10,6 +10,7 @@ lint:
     - conf/test.config
     - tests/default.config
     - tests/default.nf.test
+    - CODE_OF_CONDUCT.md
```
beatrizsavinhas, line 12: The file has been removed since the organisation is not nf-core (included in new template) but nf-core linting still expects the file to exist.

.nf-core.yml
```diff
@@ -10,7 +10,9 @@ lint:
...
   files_unchanged:
+    - .github/ISSUE_TEMPLATE/config.yml
```
beatrizsavinhas, line 14: We already had a different config, but it seems that the linting rules were updated so an error was thrown. That is why I have now added this file here.

workflows/oncorefiner.nf
```diff
@@ -60,14 +43,17 @@ workflow ONCOREFINER {
         val_genome               // string:  [optional]  genome assembly (e.g. "GRCh38")
         val_species              // string:  [optional]  species (e.g. "homo_sapiens")
         val_vep_cache_version    // string:  [optional]  version of vep cache to use (e.g. "107")
+    multiqc_config
+    multiqc_logo
+    multiqc_methods_description
+    outdir
```
beatrizsavinhas: I think the naming for these new parameters is already following the strict syntax rules, so they don't follow the naming structure we had before and don't include structure comments.

Should I modify them to follow the naming structure and alphabetical order we had before? 🤔

workflows/oncorefiner.nf
beatrizsavinhas, line 1: The template indentation, as you can see in several lines below, is different than what we had before. I kept the template changes as they come to avoid further conflicts in the future. We might want to change this style convention soon!

main.nf
```diff
@@ -27,30 +27,35 @@ include { SAMTOOLS_VIEW           } from './modules/nf-core/samtools/view/main'
...
+    val_bai_tumor                   // string:  [optional]  path to BAI file for the tumor sample
+    val_cadd_prescored_indels       // string:  [optional]  path to CADD prescored indels file
+    val_cadd_resources              // string:  [optional]  path to CADD resources directory
+    val_genome                      // string:  [optional]  genome assembly (e.g. "GRCh38")
+    val_genome_fasta                // string:  [optional]  path to genome fasta file
+    val_genome_fai                  // string:  [optional]  path to genome fasta index file
+    val_multiqc_config              // string:  [optional]  path to multiqc config file
+    val_multiqc_logo                // string:  [optional]  path to image file to be used as logo in multiqc report
+    val_multiqc_methods_description // string:  [optional]  path to text file containing methods description to be included in multiqc report
+    val_outdir                      // string:  [mandatory] path to output directory (default: ./results)
```
beatrizsavinhas, line 47: These inputs were now added since they are now required in `workflows/oncorefiner.nf`.
The rest of the changes here are merely for indentation.

CODE_OF_CONDUCT.md
beatrizsavinhas, line 1: I'm guessing CODE_OF_CONDUCT.md was removed with the template update since the organisation is not set to nf-core.

workflows/oncorefiner.nf
mathiasbio: Do we know which indentation strategy we're going for? I guess it makes sense to follow nf-core if we're going to keep updating the template. 

I don't have any strong opinions here, but I would be fine with changing the indentations of the rest of the oncorefiner.nf inputs here (similar to what you did in the main.nf I think)

nf-test.config
```diff
@@ -1,21 +1,35 @@
 config {
     // location for all nf-test tests
-    testsDir "."
+    testsDir = "."
```
mathiasbio, line 3: I like it!

workflows/oncorefiner.nf
beatrizsavinhas: The template goes:
```shell
workflow NAME {

    take:
    input_1
    input_2
```

And we currently have:
```shell
workflow NAME {

    take:
        input_1
        input_2
```

.github/workflows/nf-test.yml
```diff
@@ -66,22 +66,24 @@ jobs:
       fail-fast: false
       matrix:
         shard: ${{ fromJson(needs.nf-test-changes.outputs.shard) }}
-        profile: [docker, singularity]
+        profile: [conda, docker, singularity]
```
fevac: will we be supporting conda?

assets/sendmail_template.txt
```diff
@@ -9,12 +9,12 @@ Content-Type: text/html; charset=utf-8
...
 --nfcoremimeboundary
-Content-Type: image/png;name="nf-core-oncorefiner_logo.png"
+Content-Type: image/png;name="Clinical-Genomics-oncorefiner_logo.png"
 Content-Transfer-Encoding: base64
 Content-ID: <nfcorepipelinelogo>
-Content-Disposition: inline; filename="nf-core-oncorefiner_logo_light.png"
+Content-Disposition: inline; filename="Clinical-Genomics-oncorefiner_logo_light.png"
 
-<% out << new File("$projectDir/assets/nf-core-oncorefiner_logo_light.png").
+<% out << new File("$projectDir/assets/Clinical-Genomics-oncorefiner_logo_light.png").
```
fevac: I don't think we have changed the logo file name. But if the template tries to do this, we should probably rename in a different PR to avoid issues with updates

workflows/oncorefiner.nf
fevac: I don't know enough but I would agree with Mathias said. Maybe @fellen31 has some thoughs?

fevac (approved):
LGTM but I don't know much about template updates, you might want to have a quick review from someone else more knowledgable

.github/workflows/nf-test.yml
kristinebilgrav: Same though! I think conda can be removed

workflows/oncorefiner.nf
fellen31: I like it like the template. I think:

```
 take:
        input_1
        input_2
```

is just unnecessary indentation which becomes less readable if you have a smaller screen. But that's just personal opinion :)

kristinebilgrav (approved):
LGTM! ⭐

workflows/oncorefiner.nf
fellen31: If you choose to diverge from the nf-core template and prefix channels with `ch_`, I think you should do it for these as well + add the `// string:  [optional]  xxx`

workflows/oncorefiner.nf
```diff
@@ -103,84 +89,63 @@ workflow ONCOREFINER {
...
-
-    emit:
-    multiqc_report = MULTIQC.out.report.toList() // channel: /path/to/multiqc_report.html
+                multiqc_logo ? file(multiqc_logo, checkIfExists: true) : [],
+                [],
+                [],
+            ]
+        }
+    )
+    emit:multiqc_report = MULTIQC.out.report.map { _meta, report -> [report] }.toList() // channel: /path/to/multiqc_report.html
```
fellen31: This should be on a separate line?

nf-test.config
beatrizsavinhas: Can't take any credit - it comes from the template! 😅

workflows/oncorefiner.nf
beatrizsavinhas: Absolutely! Fixed now!

.github/workflows/nf-test.yml
beatrizsavinhas: But we have a conda (and mamba) profile in `nextflow.config`. If the idea is to not support conda in the future, that should be applied everywhere.
Perhaps I'm not understanding the full picture, but in that case, I suggest opening an issue to remove conda support but for now keep this update since it is in line with the current codebase.

workflows/oncorefiner.nf
beatrizsavinhas: See discussion in https://github.com/Clinical-Genomics/oncorefiner/pull/100/changes#r3233140944.

.github/workflows/nf-test.yml
fellen31: In Nallo we still have the profiles in nextflow.config, but have removed mentions of conda support from the documentation. This is since only parts of the pipeline can run with conda. A user could still run parts with conda, but it's not let's say "officially" supported (since it's not tested). Having the profiles in nextflow.config also makes for less merge conflicts in template PRs.  

Agree that it could be removed elsewhere or explicitly stated. But would it be out of scope for this PR to add conda back to the tests? 🤔

workflows/oncorefiner.nf
beatrizsavinhas: Regarding the indentation: Opened an issue to update it in the future - https://github.com/Clinical-Genomics/MTP-oncoflow/issues/54.

Regarding the naming format and order for these parameters: Changed it now to match the current structure!

.github/workflows/nf-test.yml
beatrizsavinhas: OK, I was under the assumption that this was a new template addition but that is not correct - it was already there and had been manually changed in previous template updates!
Given this, I agree this it is out of scope for this PR! So I will remove it.

assets/sendmail_template.txt
beatrizsavinhas: Indeed! Reverted now.

fevac (approved):
🚀

PR #99, opened 2026-05-12
feat: Add `genmod_score_config` param

beatrizsavinhas:
### Added

- Add `genmod_score_config` param to be used by  'vcf_annotate_score_genmod' subworkflow in https://github.com/Clinical-Genomics/oncorefiner/pull/97.

PR #103, opened 2026-05-13
feat: Add SNV variant scoring

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/31.

### Added

- `tests/optional_inputs_stub.nf.test` and corresponding snapshot. Includes stub test `"No genmod_score_config with '-profile test' - stub"` to test the logic added here for this condition.

From https://github.com/Clinical-Genomics/oncorefiner/pull/99:

- `genmod_score_config` param to be used by `vcf_annotate_score_genmod` subworkflow.

From https://github.com/Clinical-Genomics/oncorefiner/pull/97:

- `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow and
included modules:
	- `genmod/annotate`
	- `genmod/compound` 
	- `genmod/models`
	- `genmod/score`
- `conf/subworkflows/vcf_annotate_score_genmod.config` with the
following settings:
   - `GENMOD_SCORE`
		- `--family_id ${meta.id}` - so that the score annotation in the output
		file shows `RankScore=<meta.id>:<score>`.
		- `--rank_results` - that provides an additional `RankResult` field in
		the output vcf file which has a score breakdown by category.
    - `BCFTOOLS_VIEW`
       - `--output-type z` - to produce a zipped output vcf file.
       - `--write-index tbi` - to produces an index `tbi` file.

### Changed

- `PROCESS_SNVS` output logic to return `ch_research_filtered_vcf/tbi` after scoring if this step is run. The file names for these outputs are then given by `VCF_ANNOTATE_SCORE_GENMOD:BCFTOOLS_VIEW`, i.e. with prefix `${meta.id}_genmod_score` as specified in the config file.

From https://github.com/Clinical-Genomics/oncorefiner/pull/97:

- `nf-test.config` to also ignore tests from
`subworkflows/genomic-medicine-sweden`.
- Updated `bcftools/view` module included in `vcf_annotate_score_genmod`
subworkflow.

From https://github.com/Clinical-Genomics/oncorefiner/pull/106:

- Updated `conf/test_base.config` and
`subworkflows/local/process_snvs/tests/main.nf.test` to include the new
test `genmod_score_config` input.
- Updated test snapshot to reflect the feature changes.

From https://github.com/Clinical-Genomics/oncorefiner/pull/109:

- Updated documentation to reflect the feature changes - citations and output information.


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#59](https://github.com/Clinical-Genomics/oncorefiner/pull/59) Added `ANNOTATE_CADD` subworkflow with following test (stub only), for CADD scoring of InDels, used in `PROCESS_SNVS`.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Added `tumor_normal` config file, used by the default test profile.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Added `tumor_only` config file, profile and pipeline test and snapshot.
+- [#99](https://github.com/Clinical-Genomics/oncorefiner/pull/99) Added `genmod_score_config` parameter.
+- [#97](https://github.com/Clinical-Genomics/oncorefiner/pull/97) Added `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow to `PROCESS_SNVS`.
```
```suggestion
- [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genmod_score_config` parameter.
- [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow to `PROCESS_SNVS`.
```

CHANGELOG.md
beatrizsavinhas, line 1: This is the PR that is actually changing dev (not the other PRs for this feature branch).

conf/subworkflows/vcf_annotate_score_genmod.config
```diff
@@ -0,0 +1,32 @@
...
+        ext.prefix = File name prefix for output files.
+----------------------------------------------------------------------------------------
+*/
+
+process {
+
+    withName:'.*:VCF_ANNOTATE_SCORE_GENMOD:GENMOD_SCORE' {
+        ext.args = { [
+            "--family_id ${meta.id}",
+            "--rank_results"
```
fellen31: I'm a fan of the trailing comma.

```suggestion
            "--rank_results",
```

conf/subworkflows/vcf_annotate_score_genmod.config
```diff
@@ -0,0 +1,32 @@
...
+        ].join(' ') }
+        publishDir = [
+            enabled: false,
```
fellen31: You are a fan of the trailing here.

conf/subworkflows/vcf_annotate_score_genmod.config
```diff
@@ -0,0 +1,32 @@
...
+        ]
+   }
+
+    withName: '.*:VCF_ANNOTATE_SCORE_GENMOD:BCFTOOLS_VIEW' {
+        ext.args = { [
+            "--output-type z",
+            "--write-index=tbi"
```
fellen31:
```suggestion
            "--write-index=tbi",
```

docs/output.md
```diff
@@ -12,6 +12,7 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 <!-- TODO: Add step description here similarly to what was added in `README.md`. -->
 
+- [`Genmod score`](https://github.com/Clinical-Genomics/genmod) - Rank variants and annotate output vcf file with score infomation.
```
fellen31:
```suggestion
- [`Genmod score`](https://github.com/Clinical-Genomics/genmod) - Rank variants and annotate output VCF file with score infomation. 
```

docs/output.md
```diff
@@ -30,6 +31,27 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
...
+<details markdown="1">
+<summary>Output files</summary>
+
+- `bcftools/`
+    - `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
+    - `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
+
+</details>
+
+[`Genmod score`](https://github.com/Clinical-Genomics/genmod) assigns a score to each variant based on the genmod score config file provided. The output vcf file is annotated with the following INFO fields which reflect the assigned score:
```
fellen31:
```suggestion
[`Genmod score`](https://github.com/Clinical-Genomics/genmod) assigns a score to each variant based on the genmod score config file provided. The output VCF file is annotated with the following INFO fields which reflect the assigned score:
```

docs/output.md
```diff
@@ -30,6 +31,27 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
...
+
+```
+RankScore
+RankScoreNormalized
+RankScoreMinMax
+RankResult
+```
+
```
fellen31, line 53: Maybe 
```suggestion
```

docs/parameters.md
```diff
@@ -47,6 +47,7 @@ Annotation related files and options required for the workflow.
 | `vcfanno_toml` | Path to the vcfanno toml file. <details><summary>Help</summary><small>If no toml is passed, default configurations will be used according to genome build within the context of the pipeline.</small></details>| `string` |  |  |  |
 | `vcfanno_lua` | Path to the vcfanno lua file. <details><summary>Help</summary><small>Custom operations file (lua). For use when the built-in ops don't supply the needed reduction.</small></details>| `string` |  |  |  |
 | `svdb_query_dbs` | Databases used for structural variant annotation in vcf format. <details><summary>Help</summary><small>Path to comma-separated file containing information about the databases used for structural variant annotation.</small></details>| `string` |  |  |  |
+| `genmod_score_config` | Path to genmod score configuration file (rank model). | `string` |  |  |  |
```
fellen31, line 55: Linking to an example file could be useful.

subworkflows/local/process_snvs/main.nf
```diff
@@ -99,13 +106,25 @@ workflow PROCESS_SNVS {
...
+                vcf: tuple(meta_vcf, vcf)
+                score_config: tuple(meta_vcf, genmod_score_config)
+            }
+
+        VCF_ANNOTATE_SCORE_GENMOD (
+            ch_annotate_score_genmod_in.vcf,
+            channel.empty(),
+            channel.empty(),
+            ch_annotate_score_genmod_in.score_config,
+            true
```
fellen31: Maybe 
```suggestion
            true,
```

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#59](https://github.com/Clinical-Genomics/oncorefiner/pull/59) Added `ANNOTATE_CADD` subworkflow with following test (stub only), for CADD scoring of InDels, used in `PROCESS_SNVS`.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Added `tumor_normal` config file, used by the default test profile.
 - [#69](https://github.com/Clinical-Genomics/oncorefiner/pull/69) Added `tumor_only` config file, profile and pipeline test and snapshot.
+- [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genmod_score_config` parameter.
+- [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow to `PROCESS_SNVS`.
```
fellen31, line 36: Do you also want to add that you updated `BCFTOOLS_VIEW_RESEARCH`/`BCFTOOLS_VIEW_CLINICAL` versions?

subworkflows/local/process_snvs/main.nf
fellen31: With this implementation `BCFTOOLS_VIEW_CLINICAL` is not reached when a genmod score config is not provided as input, and therefore you will have no clinical filtering output files. I think this might not be the idea, since you may still want to apply clinical filters based on the CADD and VEP annotation.

I think channel.empty() behavior is not always very intuitive, so I would prefer a skip_genmod_score parameter to this workflow that is true whenever you don't input a genmod score file to the pipeline and wrap the `VCF_ANNOTATE_SCORE_GENMOD` block in an if statement. Alternatively, always require a genmod score file when you run the pipeline.

fellen31 (changes requested):
Nice work! I think most of it looks good, but I have one major comment on the current implementation. 

Would also suggest that you add a stub test where you don't input the genmod score file (if that is to be supported) so you can test the behavior of the pipeline in that mode.

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: Really good point about the incorrect default skipping of `BCFTOOLS_VIEW_CLINICAL`! I will update the logic to allow for clinical filtering even when the scoring is skipped!

About the way of skipping, I say that should be decided and refactored for all relevant processes in https://github.com/Clinical-Genomics/MTP-oncoflow/issues/43!
I agree with your point though! I can add that logic now until we implement the skip flags that are planned for the issue above. I think it will make it easier to refactor anyway!

conf/subworkflows/vcf_annotate_score_genmod.config
beatrizsavinhas: Yeah, let's do it! Thanks for poiting it out!

docs/parameters.md
beatrizsavinhas: Do you think looking at the test file from the test config would be a reasonable way to get an example?

subworkflows/local/process_snvs/main.nf
kristinebilgrav: Good catch!! I agree: would be nice to update the logic to allow for clinical filtering now, and then the skip functionality can be implemented in the above issue 😄

main.nf
```diff
@@ -139,6 +140,14 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
                                                  : channel.empty()
 
+    // Input for genmod_score
+    if (val_genmod_score_config) {
+        ch_genmod_score_config = channel.fromPath(val_genmod_score_config).map { it -> [[id:it.simpleName], it] }.collect()
+        val_skip_genmod_score = false
+    } else {
+        ch_genmod_score_config = channel.empty()
+        val_skip_genmod_score = true
+    }
```
beatrizsavinhas, line 137: What do you think of this logic for creating `val_skip_genmod_score` @fellen31?

beatrizsavinhas: Would also welcome your inputs about the use of `def` or not!

CHANGELOG.md
beatrizsavinhas: I agree it's relevant! Added now!

tests/tumor_normal.nf.test.snap
```diff
@@ -87,6 +96,10 @@
                 "tumor.cgh:md5,2dfa8fdd26448127347045837c7cbf7c"
             ],
             [
+                [
+                    "subject_a.vcf.gz",
```
kristinebilgrav: A bit undescriptive name. But maybe you think we will address this in https://github.com/Clinical-Genomics/CancerBioInfo/issues/100?

beatrizsavinhas: Yes, exactly! I think it's easier to fix for all files once everything is fully decided.

modules.json
```diff
@@ -22,8 +33,8 @@
                     },
                     "bcftools/view": {
                         "branch": "master",
-                        "git_sha": "6383d8fe58f9498eecd5aa303e71a4a932d1e9f6",
-                        "installed_by": ["modules"]
+                        "git_sha": "6d46786420b4d7bc88eba026eb389c0c5535d120",
+                        "installed_by": ["modules", "vcf_annotate_score_genmod"]
```
kristinebilgrav, line 37: Not fully understanding what this does, but assuming its since the subworkflow uses this module, but its not installed by the subworkflow?

beatrizsavinhas: It means that this module was both installed as a module by itself, but it was also installed by the subworkflow.
This is an automatic done when installing the subworkflow!

kristinebilgrav: Ahh, ok, thanks!

subworkflows/local/process_snvs/main.nf
```diff
@@ -99,13 +107,31 @@ workflow PROCESS_SNVS {
...
+            )
+
+            ch_clinical_filtering_in = VCF_ANNOTATE_SCORE_GENMOD.out.vcf
+                .join(VCF_ANNOTATE_SCORE_GENMOD.out.index)
+
+        } else {
+            ch_clinical_filtering_in = ENSEMBLVEP_VEP.out.vcf
+                .join(ENSEMBLVEP_VEP.out.tbi)
+        }
+
```
beatrizsavinhas: Do you agree with this logic to address your comment about running `BCFTOOLS_VIEW_CLINICAL` even when scoring is not run @fellen31 ?

docs/parameters.md
fellen31: Yeah, or maybe the one you put it reference-files? I just remember it being hard to know what each file was supposed to look like in the beginning.

subworkflows/local/process_snvs/main.nf
fellen31: > About the way of skipping, I say that should be decided and refactored for all relevant processes in [Clinical-Genomics/MTP-oncoflow#43](https://github.com/Clinical-Genomics/MTP-oncoflow/issues/43)! I agree with your point though! I can add that logic now until we implement the skip flags that are planned for the issue above. I think it will make it easier to refactor anyway!

I don't really see why you couldn't start implementing the skip as decided in https://github.com/Clinical-Genomics/MTP-oncoflow/issues/43 for in this PR. Now that the issue has been refined and you decided to go with separate `--skip_*` flags you should be able to implement this isolation to the other workflows? Will it not be an integral part when implementing future workflows after you have implemented everything in https://github.com/Clinical-Genomics/MTP-oncoflow/issues/43?

I would vote either implement it properly, or do not make this part skippable until then.

main.nf
fellen31: https://github.com/Clinical-Genomics/oncorefiner/pull/103#discussion_r3322801626

subworkflows/local/process_snvs/main.nf
```diff
@@ -38,6 +45,7 @@ workflow PROCESS_SNVS {
         ch_vep_extra_files       // channel: [optional]  [path(plugin_file1), path(plugin_file2), ...]
         val_cadd_resources       // string:  [optional]  path to CADD resources directory
         val_genome               // string:  [optional]  genome assembly (e.g. "GRCh38")
+        val_skip_genmod_score    // boolean: [mandatory] whether to skip VCF_ANNOTATE_SCORE_GENMOD process
```
fellen31: I would revert this parameter, so make it `val_run_genmod_score` instead so you can skip the negation in the if statement!

subworkflows/local/process_snvs/main.nf
fellen31: Yeah think this looks perfect! Just https://github.com/Clinical-Genomics/oncorefiner/pull/103/changes/BASE..0727f0af657c2a6c7005102179f0b1c31c057830#r3322828185

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: I torn... Not sure if I prefer that the val follows the name of the (future) flag instead of hiding the revert logic somewhere upstream in the code. 🤔

fellen31: Thinking of reverting every skip in nallo to run in the main workflow.

fellen31: I just think `if (val_run_genmod_score)` is easier to read than `if (!val_skip_genmod_score)`

beatrizsavinhas: I agree! Changed to `val_run_genmod_score`!

subworkflows/local/process_snvs/main.nf
beatrizsavinhas: Adding a skip flag is a whole feature to solve a different problem than what originated this issue. I agree it is closely related, but I'd prefer to not add yet another to this already large PR. I think the conditional logic to run `VCF_ANNOTATE_SCORE_GENMOD` or not makes sense here since it was never the intention to not run `BCFTOOLS_VIEW_CLINICAL` if the genmod score config is not provided as input.

docs/parameters.md
beatrizsavinhas: Nothing has been added to `reference-files` yet as we decided to test the score config before formally adding it. So for now we only have the simple version for testing in `test-datasets` and I think that should be enough to give an idea of the structure of the file!

subworkflows/local/process_snvs/main.nf
fellen31: Let's agree to disagree!

tests/optional_inputs_stub.nf.test
beatrizsavinhas, line 1: What do you think of this implementation @fellen31 ?

tests/optional_inputs_stub.nf.test
```diff
@@ -0,0 +1,34 @@
+nextflow_pipeline {
+
+    name "Test pipeline ONCOREFINER without providing optional inputs - stub"
```
fellen31, line 3: Sure, now the only optional input is genmod score, not sure you can/should test all of them in the same test later though.

tests/optional_inputs_stub.nf.test
```diff
@@ -0,0 +1,34 @@
...
+    script "../main.nf"
+    tag "pipeline"
+    tag "conditional_inputs_stub"
+    tag "tumor_normal"
```
fellen31: Is this tag still valid?

tests/optional_inputs_stub.nf.test
```diff
@@ -0,0 +1,34 @@
...
+                genmod_score_config = null
+            }
+        }
+
+        then {
+            // stable_name: All files + folders in ${params.outdir}/ with a stable name
+            def stable_name = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['pipeline_info/*.{html,json,txt}'])
+            assertAll(
+                { assert workflow.success},
+                { assert snapshot(
```
fellen31, line 28: You could also assert that GENMODE_SCORE is not in versions to make sure it hasn't been run.

fellen31 (commented):
Minor suggestions!

tests/optional_inputs_stub.nf.test
beatrizsavinhas: Well, it's run with tumor normal data. 🤔

fellen31: Ok! Thinking that if you run with `--tag  tumor_normal`, you might want to just run the tumor normal test, not this stub.

beatrizsavinhas: That's true! Will remove!

tests/optional_inputs_stub.nf.test
beatrizsavinhas: Do you mean in the same file? I don't really see the issue for now, but maybe I'm missing something? 🤔

tests/optional_inputs_stub.nf.test
beatrizsavinhas: Great idea! Added now!

tests/optional_inputs_stub.nf.test
beatrizsavinhas: Answered in review comments.

tests/optional_inputs_stub.nf.test
fellen31: I think I mean't the naming "without providing optional inputs" should perhaps be "without providing _genmod score_ optional inputs". But this is /nit!

beatrizsavinhas: So the name of the test in particular is already:
```
test("No `genmod_score_config` with '-profile test' - stub") {
```
Does that sound good enough to you? :)

fellen31: Yeah I totally misunderstood, I though this was the test name!

fellen31 (approved):
LGTM! 🌟 Nice work!

kristinebilgrav (approved):
LGTM! 🌟

PR #104, opened 2026-05-19
feat: Add parameters for cnv report

fevac:
Part of https://github.com/Clinical-Genomics/MTP-oncoflow/issues/34

### Added

- Parameters for cnv report: cnv_gene_tsv, cnv_segment_tsv


nextflow.config
```diff
@@ -38,6 +38,11 @@ params {
     vcfanno_extra               = null
     vcfanno_lua                 = null
 
+    // CNV report
+    cnv_gene                    = null
+    cnv_segment                 = null
+    cnv_report                  = 'cnv_report.html'
```
fellen31: Does this have to be a parameter, or could it be decided by the pipeline? Will you have parameters that lets the users decide the filename for other output files?

nextflow_schema.json
```diff
@@ -82,6 +82,31 @@
...
+                    "help_text": "This file should be a tab-separated file with a header row. Typically *.purple.cnv.somatic.tsv file from Oncoanalyser."
+                },
+                "cnv_segment": {
+                    "type": "string",
+                    "format": "file-path",
+                    "exists": true,
+                    "description": "Path to a file containing segment-level information of CNV calls.",
+                    "fa_icon": "fas fa-file-csv",
+                    "pattern": "^\\S+\\.tsv$",
+                    "help_text": "This file should be a tab-separated file with a header row. Typically *.purple.cnv.gene.tsv file from Oncoanalyser."
```
fellen31: So the oncoanalyser "gene" file goes into `segment`? 🤔

nextflow.config
```diff
@@ -38,6 +38,11 @@ params {
     vcfanno_extra               = null
     vcfanno_lua                 = null
 
+    // CNV report
+    cnv_gene                    = null
+    cnv_segment                 = null
```
fellen31: For me, these names are not very descriptive. It's not clear if this is a string to take a gene for example, or a file... It will get better with the type annotations, but until then, I wonder if it would make sense to include for example "purple" if these files are from that software, and "tsv" to indicate that this is supposed to be a tsv file?

nextflow_schema.json
```diff
@@ -82,6 +82,31 @@
...
                     "description": "Sex of the patient."
                 },
+                "cnv_gene": {
+                    "type": "string",
+                    "format": "file-path",
+                    "exists": true,
+                    "description": "Path to a file containing gene-level information of CNV calls.",
+                    "fa_icon": "fas fa-file-csv",
+                    "pattern": "^\\S+\\.tsv$",
+                    "help_text": "This file should be a tab-separated file with a header row. Typically *.purple.cnv.somatic.tsv file from Oncoanalyser."
```
fellen31: Great help text 🥇

nextflow_schema.json
fellen31: 👍

nextflow_schema.json
```diff
@@ -82,6 +82,31 @@
                     "enum": ["female", "male", "unknown"],
                     "description": "Sex of the patient."
                 },
+                "cnv_gene": {
```
kristinebilgrav: Maybe this name and the `cnv_segment` name could be adjusted to make it more intuitive that they are files? ex: `cnv_gene_tsv` and `cnv_segment_tsv` ?

fellen31 (commented):
Nice! Some comments for someone that knows nothing about the oncoanalyser output files.

kristinebilgrav (commented):
Looks good! Should these parameters already be included in the test config as well or will that be done later?

nextflow.config
beatrizsavinhas: Yeah, I don't think this should be a parameter unless there is a specific reason that I'm not seeing. All other output file names will be decided by the pipeline as per https://github.com/Clinical-Genomics/CancerBioInfo/issues/100.

beatrizsavinhas:
> Looks good! Should these parameters already be included in the test config as well or will that be done later?

I agree with this line of thought! Since this is a standalone PR, it would be nice to add the parameters to the test config files already as this is also an opportunity to test if the patterns are correct because there is no other test here for this code.

nextflow.config
fevac: yeah, excellent point. removed!

nextflow_schema.json
fevac: good catch! they are mixed up!

fevac: fixed

nextflow.config
fevac: fair enough. I'm not a big fan of type information in the variable names (specially as a python programmer where you would use type hints instead). I do see the issue here. However, isn't it enough with adding explanations about expected input in the nextflow_schema.json?

nextflow_schema.json
fevac: I'm ok with changing if this is the general feeling, however see https://github.com/Clinical-Genomics/oncorefiner/pull/104/changes#r3303942895

nextflow.config
fevac: changed to:
    cnv_gene_tsv                = null
    cnv_segment_tsv             = null

nextflow_schema.json
fevac: changed!

nextflow.config
fellen31: I see your point as well! Okay, so with type hinting you (as the developer) would see that these are Paths. Path to a `cnv_gene` file. Path to a `cnv_segment` file. Is that enough? Not sure how/if type hinting also helps the user of the pipeline.

 If they are aways going to be files from the purple tool I think it could be nice to include it in the name. The description in `nextflow_schema.json` is well written I think, but I think there's also value in having self-explanatory variable names if possible, so you don't have to go to the documentation to get a sense of what they are.

fevac:
> Looks good! Should these parameters already be included in the test config as well or will that be done later?

This will be done when adding the test dataset

kristinebilgrav (approved):
Great! 🌟

PR #105, opened 2026-05-20
Subworkflow specific configs

kristinebilgrav:
Closes https://github.com/Clinical-Genomics/CancerBioInfo/issues/144

### Added

- subworkflow specific configs for tests


subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -35,16 +36,15 @@ nextflow_workflow {
...
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
                     // All directories and files in the output directory
                     output_directories_and_files,
-                    // All files in the output directory with their md5 checksum
-                    output_files.collect { file -> [ file.getName(), path(file.toString()).md5 ] }
+                    // Output file with md5 checksum
+                    file(workflow.out.cgh.get(0).get(1)).name,
+                    path(workflow.out.cgh.get(0).get(1)).md5
```
kristinebilgrav: Might want to change to this with new publishing strategy? https://github.com/Clinical-Genomics/CancerBioInfo/issues/112

subworkflows/local/generate_cytosure_files/tests/nextflow.config
```diff
@@ -0,0 +1,32 @@
+process {
+
+    withName: 'TIDDIT_COV' {
+        publishDir = [
+            enabled: false,
```
kristinebilgrav: Wait to https://github.com/Clinical-Genomics/CancerBioInfo/issues/112 and subsequent are merged to avoid this?

subworkflows/local/generate_cytosure_files/tests/main.nf.test
fellen31: If you change, I would also delete `output_directories_and_files` :)

subworkflows/local/process_snvs/tests/main.nf.test.snap
```diff
@@ -70,10 +70,10 @@
                 ]
             ]
         ],
-        "timestamp": "2026-06-15T16:01:56.128354",
+        "timestamp": "2026-08-05T13:48:42.074485",
         "meta": {
             "nf-test": "0.9.5",
-            "nextflow": "26.04.2"
+            "nextflow": "26.04.6"
```
mathiasbio, line 76: What are these nextflow version changes? Just depends on the local version of nextflow when running the test?

kristinebilgrav: Exactly, so they dont really matter. I guess these are actually a bit unnecessary since there are no actual changes to the snapshot. Let me know if you want me to revert back to the old versions.

mathiasbio: I don't think it matters, just curious! :)

PR #106, opened 2026-05-22
fix: Update tests and snapshots

beatrizsavinhas:
### Fixed

- Updated `conf/test_base.config` and `subworkflows/local/process_snvs/tests/main.nf.test` to include the new test `genmod_score_config` input.
- Updated test snapshot to reflect the feature changes.

------

:x: Changelog will be updated in feature branch in https://github.com/Clinical-Genomics/oncorefiner/pull/103.


fellen31 (approved):
Find it a bit tricky to review the changes in the snapshots in this PR, when I need to look up another PR for what actually caused the snapshots to change..

But maybe I'm just not used to it and will come around!

beatrizsavinhas:
> Find it a bit tricky to review the changes in the snapshots in this PR, when I need to look up another PR for what actually caused the snapshots to change..
> 
> But maybe I'm just not used to it and will come around!

Humm I can understand your struggle! One definitely needs to either be aware of what is in the feature branch already or look at the main PR - https://github.com/Clinical-Genomics/oncorefiner/pull/103.
The alternative would be to lump together the test/snapshot updates with the other changes, but that would have made a quite large PR. I thought that this could be easier to review, but am totally open to other ways of doing this! Let me know if you have any ideas. :)

kristinebilgrav (approved):
🌟 

I agree with above, do need to go back to see what was done to understand. But dont know what is best either, so 🤷‍♀️

PR #109, opened 2026-05-25
docs: Update documentation

beatrizsavinhas:
Part of https://github.com/Clinical-Genomics/oncorefiner/pull/103.

### Changed

- Updated documentation to reflect the feature changes - citations and output information.

------

:x: Changelog will be updated in feature branch in https://github.com/Clinical-Genomics/oncorefiner/pull/103.

### DIFF
--- a/docs/output.md
+++ b/docs/output.md
@@ -41,6 +41,9 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 </details>
 
+[`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
+
+
 ### MultiQC
 
 <details markdown="1">
--- a/docs/output.md
+++ b/docs/output.md
@@ -12,6 +12,7 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 <!-- TODO: Add step description here similarly to what was added in `README.md`. -->
 
+- [`Genmod score`](https://github.com/Clinical-Genomics/genmod) - Rank variants and annotate output vcf file with score infomation. 
 - [MultiQC](#multiqc) - Aggregate report describing results and QC from the whole pipeline
 - [Pipeline information](#pipeline-information) - Report metrics generated during the workflow execution
 
@@ -30,7 +31,7 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 [`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
 
-### `PROCESS_SNVS:VCF_ANNOTATE_SCORE_GENMOD`
+### `Genmod score`
 
 <details markdown="1">
 <summary>Output files</summary>
--- a/docs/output.md
+++ b/docs/output.md
@@ -35,7 +35,7 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 <details markdown="1">
 <summary>Output files</summary>
 
-- `bcftools``
+- `bcftools`
 - `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
 - `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
 
--- a/docs/output.md
+++ b/docs/output.md
@@ -41,7 +41,14 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 </details>
 
-[`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
+[`Genmod score`](https://github.com/Clinical-Genomics/genmod) assigns a score to each variant based on the genmod score config file provided. The output vcf file is annotated with the following INFO fields which reflect the assigned score:
+
+```
+RankScore
+RankScoreNormalized
+RankScoreMinMax
+RankResult
+```
 
 
 ### MultiQC
--- a/docs/output.md
+++ b/docs/output.md
@@ -35,9 +35,9 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 <details markdown="1">
 <summary>Output files</summary>
 
-- `bcftools`
-- `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
-- `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
+- `bcftools/`
+    - `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
+    - `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
 
 </details>
 
--- a/docs/output.md
+++ b/docs/output.md
@@ -30,6 +30,17 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 [`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
 
+### `PROCESS_SNVS:VCF_ANNOTATE_SCORE_GENMOD`
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `bcftools``
+- `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
+- `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
+
+</details>
+
 ### MultiQC
 
 <details markdown="1">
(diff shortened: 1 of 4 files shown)


docs/output.md
```diff
@@ -30,6 +30,17 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
...
 [`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
 
+### `PROCESS_SNVS:VCF_ANNOTATE_SCORE_GENMOD`
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `bcftools`
+- `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
+- `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
```
fellen31: Think this is missing:

```suggestion
- `bcftools/`
    - `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
    - `<meta.id>.vcf.gz.tbi`: index file for the gzipped VCF file.
```

docs/output.md
```diff
@@ -30,6 +30,17 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
 
 [`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
 
+### `PROCESS_SNVS:VCF_ANNOTATE_SCORE_GENMOD`
```
fellen31: Would put "Ranked variants"/"Variant ranking" which is simpler for someone reading the documentation that may not be very familiar with the pipeline.  

```suggestion
### Ranked variants
```

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
```diff
@@ -204,6 +205,12 @@ def toolCitationText() {
             svdb           +
             bcftools_view  +
             ensemblvep_vep
+
+        if (params.genmod_score_config) {
+            citations_list =
+                citations_list +
+                genmod
+        }
```
fellen31: Should this not be in the SNV section? I think I would prefer it being it's own if block for simplicity, i.e. `if (params.sv_vcf && params.genmod_score_config)`.

fellen31 (commented):
Cool! 

Consider if you want to have a paragraph similar to the one above for Vcfanno: 

```
[`Vcfanno`](https://github.com/brentp/vcfanno) annotates VCF files with a number of INFO fields from the VCFs or BED files provided.
```

I'm not a fan of duplication in the output and usage documentations, but the section might look a bit empty without, especially as you've hidden the output files:

Image

docs/output.md
kristinebilgrav: Or alternatively `GENMOD`

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
```diff
@@ -204,6 +205,12 @@ def toolCitationText() {
             svdb           +
             bcftools_view  +
             ensemblvep_vep
+
+        if (params.genmod_score_config) {
+            citations_list =
+                citations_list +
+                genmod
```
kristinebilgrav: Why is it not added with the rest above? Because its not always run or?

docs/output.md
```diff
@@ -30,6 +30,17 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
...
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `bcftools`
+- `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs.
```
fevac: Maybe worth mentioning the info field where the scoring is added

```suggestion
- `<meta.id>.vcf.gz`: a gzipped VCF file containing ranked and annotated SNVs. The INFO field includes `RankScore`, which stores the variant ranking score.
```

docs/output.md
fevac: Agree. Ranked variants is easier to understand. You could add  a parenthesis or an additional line indicating the tool


```suggestion
### Ranked variants (Genmod)
```

beatrizsavinhas: The thing is that there will be a different "ranked variants" for the SV flow in the future. How do you suggest distinguishing between the two?

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
beatrizsavinhas: Yes, it's only run if the `genmod_score_config` is given.

docs/output.md
beatrizsavinhas: Yes! Thank you!

subworkflows/local/utils_nfcore_oncorefiner_pipeline/main.nf
beatrizsavinhas: It should absolutely be in the SNV section! I was distracted. 😅 

About the block division, I actually think the current solution is more readable as it's easy to see that this logic is part of the SNV section.

docs/output.md
fellen31: I think you could use subheadings :) May not be the best approach, but an example at least: https://genomic-medicine-sweden.github.io/nallo/latest/output/#ranking

docs/output.md
beatrizsavinhas: All of these INFO fields are added:

```
##INFO=<ID=RankScore,Number=.,Type=String,Description="The rank score for this variant in this family. family_id:rank_score.">
##INFO=<ID=RankScoreNormalized,Number=.,Type=String,Description="The normalized rank score in range(0, 1) for this variant in this family. family_id:rank_score.">
##INFO=<ID=RankScoreMinMax,Number=.,Type=String,Description="The rank score MIN-MAX bounds. family_id:min:max.">
##INFO=<ID=RankResult,Number=.,Type=String,Description="clinical_significance">
```

Should we add information about all of them?

docs/output.md
```diff
@@ -30,6 +30,27 @@ The pipeline is built using [Nextflow](https://www.nextflow.io/) and processes d
...
+</details>
+
+[`Genmod score`](https://github.com/Clinical-Genomics/genmod) assigns a score to each variant based on the genmod score config file provided. The output vcf file is annotated with the following INFO fields which reflect the assigned score:
+
+```
+RankScore
+RankScoreNormalized
+RankScoreMinMax
+RankResult
+```
```
beatrizsavinhas, line 52: @fellen31 Thank for pointing out that the description was missing! I agree and also dislike the information duplication in the output and usage documentations, so we should think about how to better structure this in the future. But for now, since this is the format in place, I think it makes sense to add this here too!

@fevac Is this information now sufficient to address your previous comment https://github.com/Clinical-Genomics/oncorefiner/pull/109#discussion_r3302220494?

docs/output.md
beatrizsavinhas: See https://github.com/Clinical-Genomics/oncorefiner/pull/109/changes#r3302486772.

docs/output.md
beatrizsavinhas: Alright, so the current logic of the document is that tools are listed. To follow that, I renamed the section `Genmod score` since that is the most relevant tool. It is confusing that the output directory stated in this section is `bcftools`, but this will change and should be addressed in https://github.com/Clinical-Genomics/CancerBioInfo/issues/100.
I find the current structure of the document (which comes from the nf-core template) confusing. I think as a user I would either want to have a clear list of all output directories/files and what they contain or a clear connection between the processes and their outputs (which is why I had hinted at the process name initially). But this is a discussion for another issue so for now I propose simply following the structure that is already there.

Regarding the subheadings, I think this would be a good idea! Heads up @kristinebilgrav to think about this when https://github.com/Clinical-Genomics/MTP-oncoflow/issues/32 is implemented.

fellen31: >  But this is a discussion for another issue so for now I propose simply following the structure that is already there.

Sounds okay. Not sure I know yet how beneficial I think it is to just keep going with a sub-optimal solution. But maybe consistently bad is better than inconsistency 😅

docs/output.md
fevac: yes thanks! 🌟

fellen31: Not saying this is wrong, but do would you also add info on info fields added by vcfanno or VEP then?

fellen31 (approved):
Looks okay to me. There's no update to the usage documentation though, but maybe that will be done in another PR?

fevac (approved):
🚀

beatrizsavinhas:
> Looks okay to me. There's no update to the usage documentation though, but maybe that will be done in another PR?

No, I didn't think it was relevant to add anything to the usage documentation since that mostly contains information on how to start a run which this feature doesn't change.

docs/output.md
beatrizsavinhas: Also thought of that! I think we will have to decide in the future how specific we want to be in this document. But this file is currently not updated at all and the information is inconsistent.

Maybe all of these discussions could be addressed in https://github.com/Clinical-Genomics/MTP-oncoflow/issues/44 so I added a comment there!

docs/output.md
beatrizsavinhas: See https://github.com/Clinical-Genomics/oncorefiner/pull/109#discussion_r3302935816!

PR #110, opened 2026-05-27
fix: Add genome parameter for `VCF2CYTOSURE`

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/68.

### Added

- Argument `genome` for `VCF2CYTOSURE`, given by `params.genome_version_number`, to ensure the reference genome used for the process matches the given input data.

### DIFF
--- a/conf/subworkflows/generate_cytosure_files.config
+++ b/conf/subworkflows/generate_cytosure_files.config
@@ -34,7 +34,7 @@ process {
             args = { [
                 "--bins 20",
                 "--sex ${params.sex}",
-                "--genome ${params.genome.replace("GRCh","")}" // Remove "GRCh" from genome name to obtain only the version number (e.g. "GRCh38" -> "38")
+                "--genome ${params.genome_version_number}"
             ].join(' ') }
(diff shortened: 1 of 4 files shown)


conf/subworkflows/generate_cytosure_files.config
```diff
@@ -33,7 +33,8 @@ process {
         ext{
             args = { [
                 "--bins 20",
-                "--sex ${params.sex}"
+                "--sex ${params.sex}",
+                "--genome ${params.genome.replaceAll("[a-zA-Z]","")}" // Remove alphabetical characters from genome name to obtain only the version number (e.g. "GRCh38" -> "38")
```
fellen31: I think if you change to this the code is simpler self explanatory enough that you don't need the comment. 

```suggestion
                "--genome ${params.genome.replace("GRCh","")}"
```

beatrizsavinhas: Ah that's a good point! Thank you!

kristinebilgrav (approved):
amazing! 🚀

PR #111, opened 2026-05-27
feat: Add hidden `genome_version_number` parameter

beatrizsavinhas:
Part of https://github.com/Clinical-Genomics/MTP-oncoflow/issues/68.

### Added

- Hidden `genome_version_number` parameter parsed from params.genome by removing "GRCh" from the genome name to obtain only the version number (e.g. "GRCh38" -> "38"). This is necessary to provide config settings to fix the bug in the issue above.

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -45,7 +45,7 @@ params {
     fasta                       = null
     fai                         = null
     genome                      = 'GRCh38'
-    genome_version_number       = params.genome.replace("GRCh","") // Remove "GRCh" from genome name to obtain only the version number (e.g. "GRCh38" -> "38")
+    genome_version_number       = params.genome.replace("GRCh","") as Integer // Remove "GRCh" from genome name to obtain only the version number (e.g. "GRCh38" -> "38")
     species                     = 'homo_sapiens'
 
     // MultiQC options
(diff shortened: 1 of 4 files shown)


kristinebilgrav (approved):
👍

nextflow.config
```diff
@@ -45,6 +45,7 @@ params {
     fasta                       = null
     fai                         = null
     genome                      = 'GRCh38'
+    genome_version_number       = params.genome.replace("GRCh","") as Integer // Remove "GRCh" from genome name to obtain only the version number (e.g. "GRCh38" -> "38")
```
fevac, line 48: why is this a parameter and not a variable?
The risk here would be that genome and genome_version_number fall out of sync. You could specify  without consequences
```
genome                                    = 'GRCh38'
genome_version_number       = 37
```
I'd suggest to make this a variable instead and only specify the genome once

beatrizsavinhas: A variable cannot be accessed by config files, which is necessary to fix this bug (see https://github.com/Clinical-Genomics/oncorefiner/pull/110).

There is indeed a risk, that this mismatch could occur. This is the case for many parameters at the moment. We could think about adding this kind of validation of input parameters in the future! 🤔

beatrizsavinhas: Opened an issue to document this - https://github.com/Clinical-Genomics/MTP-oncoflow/issues/70.

fevac: soudns good

fevac (approved):

PR #114, opened 2026-05-29
feat: add sv annotation scripts

kristinebilgrav:
Branch that will add python scripts to the branch `annotate-sv-linx` which will contain everything needed to annotate the SV vcf with LINX information
Links to main issue: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/67

### Added

- Script to merge LINX tsv files:
    - We only need the fusion names and whether its reported or not for now
- Script to annotate SV vcf with information from LINX files 
    - Can accept any tsv file with the format ID, TAG, TAG, TAG - thus will not need changing if we add more in the future


modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,98 @@
...
+        vcf_out.write(record)
+
+    vcf_in.close()
+    vcf_out.close()
+    output_index = pysam.tabix_index(output_file, preset="vcf", force=True)
+
+    return output_file, output_index
+
+# run
+new_header_dict = {}
```
kristinebilgrav: Was originally made to verify all annotation fields were mentioned in the header, but pysam will also complain if they are not - so probably will remove!

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,98 @@
...
+        try:
+            # extract the ID, Number, Type from the header line
+            id = line.split('ID=')[1].split(',')[0]
+            number = line.split('Number=')[1].split(',')[0]
+            type = line.split('Type=')[1].split(',')[0]
+            header_dict[id] = {'number': number, 'type': type}
+
+            # if the type is Flag, check that number is 0, and if not, raise an error - May be handled by pysam?
+            if type == 'Flag' and number != '0':
+                raise ValueError(f"Header line {line} has type Flag but number is not 0")
```
kristinebilgrav: Need to investigate more - but I believe pysam will complain if its not in the correct format, so this part might be unneccesary

kristinebilgrav: Can confirm that pysam throws an error if not the correct format. Will remove this part

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
kristinebilgrav: See comment above. Will remove

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,84 @@
...
+    with open(tsv_file, newline='') as f:
+        reader = csv.DictReader(f, delimiter="\t") # convert each tsv row to dict
+
+        id_column = reader.fieldnames[0]
+        for row in reader:
+            key = row.pop(id_column) # removes id column from dict, returns id (value)
+
+            # convert numeric strings to ints
+            values = { k: int(v) if v.isdigit() else v
+                      for k, v in row.items() }
```
kristinebilgrav: If multiple fusion names on the same entry, will be overwritten: to fix

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(description="Merge LINX files into one TSV file")
+
+    parser.add_argument("-f", type=str, help="The fusion linx tsv file", required=True)
+    parser.add_argument("-b", type=str, help="The breakends linx tsv file", required=True)
+    parser.add_argument("-s", type=str, help="The svs linx tsv file", required=True)
+    parser.add_argument("-o", type=str, help="name of output tsv file", required=True)
+
+    args = parser.parse_args()
```
beatrizsavinhas: I highly discourage using `parser` since the necessary inputs of the function are not explicit. Suggest using `click` instead.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # load tsv into pandas df
+    fusions   = pd.read_csv(args.f, sep='\t', dtype=str)
+    breakends = pd.read_csv(args.b, sep='\t', dtype=str)
+    svs       = pd.read_csv(args.s, sep='\t', dtype=str)
+
+    output_file = args.o
```
beatrizsavinhas: In addition to the comment above, I highly discourage using abbreviated names for variables. It is really not intuitive to read `args.f` and know what this argument is expected to hold.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+
+    merge_linx_files(fusions, breakends, svs, output_file)
```
beatrizsavinhas: If using `click`, this function is the only necessary function to call below.

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
+import pandas as pd
+import argparse
+
+"""
+This script aims to combine the three linx files (fusions, breakends, svs) into one merged tsv file
```
beatrizsavinhas, line 7: Add information on inputs and expected formats. See other scripts/tools for reference.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
+import pandas as pd
+import argparse
+
+"""
+This script aims to combine the three linx files (fusions, breakends, svs) into one merged tsv file
+"""
+
+def merge_linx_files(fusions, breakends, svs, output_file):
+
+    fusions.rename(columns={'name': 'FUSION_NAME', 'reported': 'REPORTED'}, inplace=True)
```
beatrizsavinhas: Why is this necessary?

bin/combine_linx_tsv.py
beatrizsavinhas, line 1: Add typehints!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+    fusions['REPORTED'] = fusions['REPORTED'].replace({'false': 0, 'true': 1}) # to adhere to pysam
+
+    # merge fusions and breakends on 'fivePrimeBreakendId' & 'threePrimeBreakendId' and 'id' columns
+    fusion_fivebreakend = fusions[['fivePrimeBreakendId', 'threePrimeBreakendId', 'FUSION_NAME', 'REPORTED']].merge(breakends[['id', 'svId']], left_on='fivePrimeBreakendId', right_on='id', how='left')
```
beatrizsavinhas: I find this line hard to read. It's both subsetting the `fusions` and `breakends` dataframes and then merging them. I suggest separating these steps clearly.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+    fusion_threebreakend = fusions[['fivePrimeBreakendId', 'threePrimeBreakendId', 'FUSION_NAME', 'REPORTED']].merge(breakends[['id', 'svId']], left_on='threePrimeBreakendId', right_on='id', how='left')
```
beatrizsavinhas: The subsetting once again here is code duplication and prone to typos. If the logic is atomised as suggested above, this code duplication can be avoided.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # merge the two
```
beatrizsavinhas: I don't understand the intention from reading this comment.

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,102 @@
+import sys
+import pysam
+import csv
+
+"""
+This script aims to annotate a vcf using:
+1. a VCF file following the specifications, where the ID field will be used to annoate entries
```
beatrizsavinhas: What specifications?

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,102 @@
...
+NOTE: the script adapts the annotation style to the Types in the header.
+This means that the ‘Flag’ type indicates that the INFO field does not contain a Value entry, and hence the Number should be
+0 and the flag will be added as is to the INFO field based on 0 (false) or 1 (true).
+"""
+
+# get input files from command line arguments, TODO add click
+vcf_file = sys.argv[1]
+tsv_file = sys.argv[2]
+header_file = sys.argv[3]
+output_file = sys.argv[4]
```
beatrizsavinhas: I already see your `TODO` note and yes, this is highly discouraged.

bin/annotate_vcf_by_id.py
beatrizsavinhas, line 1: Typehints missing here as well.

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,102 @@
...
+    vcf_out.close()
+    output_index = pysam.tabix_index(output_file, preset="vcf", force=True)
+
+    return output_file, output_index
+
+# run
+
+# parse new header
+new_header_lines = []
+new_header_lines = new_header(header_file, new_header_lines)
```
beatrizsavinhas: Unnecessary duplication of `new_header_lines`.

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,102 @@
...
+# parse new header
+def new_header(header_file, header_lines):
+    f= open(header_file, 'rt')
+    for line in f:
+        line = line.strip()
+
+        # append line to list to be added to new header
+        header_lines.append(line)
+
+    return header_lines
```
beatrizsavinhas: So the idea is to generate a list containing all the lines from the `header_file`? This is not understandable from reading the name of the function.

Also, if this is the idea, I see no need to reinvent the wheel with a custom function.

```suggestion
with open(header_file_path) as f:
    addional_header_lines = f.read().splitlines()
```

Or even better, given that `header_file_path` is a `pathlib.Path` (which you easily specify with `click`):

```suggestion
addional_header_lines = header_file_path.read_text().splitlines()
```

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
```diff
@@ -0,0 +1,102 @@
...
+
+# parse tsv file
+def parse_tsv(tsv_file):
```
beatrizsavinhas: I'm having a really hard time understanding the aim of this function from reading the code. Could you summarise what format you want to achieve from the `tsv` file? Why isn't this parsing done in the dedicated script to generate the tsv file?

beatrizsavinhas (changes requested):

bin/combine_linx_tsv.py
kristinebilgrav: Made a try now!

bin/combine_linx_tsv.py
kristinebilgrav: Added some information - was this what you were thinking of?

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Just to make the names more informational for the VCF. This name will be in the header of the output file, and that again should be the name of the actual flag/information we want to add. So FUSION_NAME is more informational than just name=GENE1_GENE2

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Seperated now, let me know what you think!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Good point. Let me know if the update is what you were thinking of.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Updated

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: I added click now, but its my first time using it, so let me know if there are anything I should do differently

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: I interpreted that in the way that I could have this function as my main one hehe, let me know if this was not the case 😅

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
beatrizsavinhas: In that case add a comment explaining your reasoning because it's not clear from the code that this is needed!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,64 @@
...
+- Fusion TSV file: fivePrimeBreakendId, threePrimeBreakendId, name, reported
+- Breakend TSV file: id, svId
+- SV TSV file: svId, vcfId
+"""
+
+@click.command()
+@click.option("-f", "--fusion_file", type=click.Path(exists=True), help="The fusion linx tsv file", required=True)
+@click.option("-b", "--breakend_file", type=click.Path(exists=True), help="The breakends linx tsv file", required=True)
+@click.option("-sv", "--sv_file", type=click.Path(exists=True), help="The svs linx tsv file", required=True)
+@click.option("-o", "--output_file", type=click.Path(), help="name of output tsv file", required=True)
```
beatrizsavinhas: Much better! And then you already get automatic type and additional checks i.e. `exists=True`!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,64 @@
...
+
+def merge_linx_files(fusion_file: str, breakend_file: str, sv_file: str, output_file: str) -> None:
```
beatrizsavinhas:
```suggestion
def merge_linx_files(
	fusion_file: click.Path,
	breakend_file: click.Path,
	sv_file: click.Path,
	output_file: click.Path) -> None:
```

Much better to have the necassary arguments being explicit! Just the typehints are not quite correct.

Nit: `fusion_file_path` would be a more explanatory variable name.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # load tsv into pandas df
+    fusions   = pd.read_csv(fusion_file, sep='\t', dtype=str)
```
beatrizsavinhas: Typehints:
```suggestion
    fusions: pd.DataFrame = pd.read_csv(fusion_file, sep='\t', dtype=str)
```
Or just `DataFrame´, depending on how it is imported, I think. 🤔 
Same below!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,64 @@
...
+    breakends = pd.read_csv(breakend_file, sep='\t', dtype=str)
+    svs       = pd.read_csv(sv_file, sep='\t', dtype=str)
+
+    fusions.rename(columns={'name': 'FUSION_NAME', 'reported': 'REPORTED'}, inplace=True) # rename columns to desired header names in subsequenct VCF annotation step
+    fusions['REPORTED'] = fusions['REPORTED'].replace({'false': 0, 'true': 1}) # to adhere to pysam
+
+    # subset dataframes
+    fusions = fusions[['fivePrimeBreakendId', 'threePrimeBreakendId', 'FUSION_NAME', 'REPORTED']]
+    breakends = breakends[['id', 'svId']]
```
beatrizsavinhas: Avoid overriding variables!

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
kristinebilgrav: Of course! I chose to use pysam to annotate the INFO field in the VCF. And the input for that can be a dictionary with all the TAGs and values to add (and the TAG also needs to be present in the header with the same name). And if there are multiple entries for one TAG this can be passed as a tuple. So I therefore here parse the TSV into a dict that can be given to pysam. 
So the format of the output nested dict is: `{ ID: {TAG : TAGVALUE, TAG2: VALUE } , ID: {TAG: (value, value) } }`

And then in the following function I extract the dictionary for the specific ID which matches the VCF record, and then pass that dictionary (ie `{TAG : TAGVALUE, TAG2: VALUE }` ) to pysam. 

The idea is that you can provide any TSV file with the format ID, TAG,TAG,TAG and have it annotate with all fields. And the previous script makes this file

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # remove duplicates
+    result = result.drop_duplicates()
+
+    # save the final merged dataframe to a tsv file
+    result.to_csv(output_file, sep='\t', index=False)
+
+# main
+if __name__ == "__main__":
+    merge_linx_files()
```
beatrizsavinhas: When reading this code, I do not easily understand what the script does. `merge_linx_files` has many responsibilities and only 1 of them is explicitly stated in the name.
I think it would be nice to separate the logic to make it more explicit and easily understandable:

```suggestion
if __name__ == "__main__":
	
    combine_linx_tsv()
```

```python
<click set up>
def combine_linx_tsv(<arguments>) -> None:
	# Read tsv files into pandas DataFrames
	(...)
	merged_df: pd.DataFrame = merge_linx_dataframes(fusions, breakends, svs)
	
    merged_df.to_csv(output_file_path, sep='\t', index=False)
```

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
beatrizsavinhas: Correction: `click` returns `click.Path` objects. Not sure if these have a corresponding method for `.read_text()`.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
beatrizsavinhas: Nice! I think it looks good so far! 🌟

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
beatrizsavinhas: Yes! Nice! I left another suggestion on this though so refer to my new comment.

bin/combine_linx_tsv.py
beatrizsavinhas: Yeah this is good! Could you also add information on the output?

bin/annotate_vcf_by_id.py
kristinebilgrav: Added!

modules/local/linx_annotation/linx_vcf/scripts/annotate_vcf_linx.py
kristinebilgrav: Couldnt figure out how to do the second option, but moved this down to the annotate vcf function as its only two lines now.

bin/combine_linx_tsv.py
kristinebilgrav: Yes!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Ah, okay I understand, updated now.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Ah I see!

bin/annotate_vcf_by_id.py
kristinebilgrav: Although its not entirely clear to me all the things I should add typehints to, so LMK

beatrizsavinhas: Basically everything unless it's very obvious which object type is being created i.e. `variable = MyObject(<arguments>)`.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+def combine_linx_files(
+    fusion_file: click.Path,
+    breakend_file: click.Path,
+    sv_file: click.Path,
+    output_file: click.Path) -> None:
+
+    # load tsv into pandas df
+    fusions: pd.DataFrame   = pd.read_csv(fusion_file, sep='\t', dtype=str)
+    breakends: pd.DataFrame = pd.read_csv(breakend_file, sep='\t', dtype=str)
+    svs: pd.DataFrame       = pd.read_csv(sv_file, sep='\t', dtype=str)
```
beatrizsavinhas: Funny that you still used the Harshil alignment for python code hahaha!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+- Breakend TSV file
+- SV TSV file
+
+At least the following columns are expected in the input files:
+- Fusion TSV file: fivePrimeBreakendId, threePrimeBreakendId, name, reported
+- Breakend TSV file: id, svId
+- SV TSV file: svId, vcfId
+
+The output is a merged tsv file with the following columns:
+- vcfId: the ID of the SV in the VCF file, from the svId column in the SV TSV file
```
fevac: that should be vcfId, I believe
```suggestion
- vcfId: the ID of the SV in the VCF file, from the vcfId column in the SV TSV file
```

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
+import click
+import pandas as pd
+
+"""
```
fevac, line 6: very nice and informative description! 🌟

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+
+def merged_linx_files(fusions: pd.DataFrame, breakends: pd.DataFrame, svs: pd.DataFrame) -> pd.DataFrame:
+
+    """ Merges input tsv files and returns a dataframe with the relevant columns for annotation of VCF file"""
+
+    # rename columns to desired header names in subsequenct VCF annotation step and change reported column to 0/1
+    fusions.rename(columns ={'name': 'FUSION_NAME', 'reported': 'REPORTED'}, inplace=True)
+    fusions['REPORTED'] = fusions['REPORTED'].replace({'false': 0, 'true': 1}) # to adhere to pysam
```
fevac: Better to put comments closest to where it happens as possible. Avoid having both new line comments and end of line comments and merge whenever possible.
```suggestion
    # rename columns to desired header names in subsequenct VCF annotation step 
    fusions.rename(columns ={'name': 'FUSION_NAME', 'reported': 'REPORTED'}, inplace=True)
    # change reported column to 0/1 to adhere to pysam
    fusions['REPORTED'] = fusions['REPORTED'].replace({'false': 0, 'true': 1}) 
```

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # subset dataframes
+    fusions_subset = fusions[['fivePrimeBreakendId', 'threePrimeBreakendId', 'FUSION_NAME', 'REPORTED']]
+    breakends_subset = breakends[['id', 'svId']]
+
+    # merge fusions and breakends on 'fivePrimeBreakendId' & 'threePrimeBreakendId' and 'id' columns
+    fusion_fivebreakend = fusions_subset.merge(breakends_subset, left_on='fivePrimeBreakendId', right_on='id', how='left')
+    fusion_threebreakend = fusions_subset.merge(breakends_subset, left_on='threePrimeBreakendId', right_on='id', how='left')
+
+    # merge fusion_fivebreakend and fusion_threebreakend to get all fusions
```
fevac:
```suggestion
    # concatenate fusion_fivebreakend and fusion_threebreakend to get all fusions
```

fevac: there is a fundamental distinction between concatenating and merging

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+    fusion_fivebreakend = fusions_subset.merge(breakends_subset, left_on='fivePrimeBreakendId', right_on='id', how='left')
+    fusion_threebreakend = fusions_subset.merge(breakends_subset, left_on='threePrimeBreakendId', right_on='id', how='left')
+
+    # merge fusion_fivebreakend and fusion_threebreakend to get all fusions
+    fusion_breakend_merge = pd.concat([fusion_fivebreakend, fusion_threebreakend], ignore_index=True)
+
+    # merge fusion_breakend_merge with svs on 'svId' column in sv file
+    fusion_breakend_svs_merge = fusion_breakend_merge.merge(svs, on='svId', how='left')
+
+    # keep only the relevant columns for the final output - for debug: add columns svId, fivePrimeBreakendId, threePrimeBreakendId
```
fevac, line 49: either add the columns for debugging (and update description) or remove the comment. I would add the columns, since it's going to help us in the verification of cases and debugging

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+    result = fusion_breakend_svs_merge[
+        [
+            "vcfId",
+            "FUSION_NAME",
+            "REPORTED"
+        ]
+    ]
+
+    # remove duplicates
```
fevac, line 54: are there duplicates? why?

fevac: If we are not expecting duplicates it might be better to skip this and let it fail if it happens so we could better understand why

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+At least the following columns are expected in the input files:
+- Fusion TSV file: fivePrimeBreakendId, threePrimeBreakendId, name, reported
+- Breakend TSV file: id, svId
+- SV TSV file: svId, vcfId
+
+The output is a merged tsv file with the following columns:
+- vcfId: the ID of the SV in the VCF file, from the svId column in the SV TSV file
+- FUSION_NAME: the name of the fusion, from the name column in the fusion TSV file
+- REPORTED: whether the fusion is reported, from the reported column in the fusion TSV file, converted to 0/1 for compatibility with pysam specifications
+
```
fevac:
```suggestion
- svId, fivePrimeBreakendId, threePrimeBreakendId: Linx IDs for SVs, 5' and 3' breakends. Added for debugging purpose.
- 

```

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
+import click
+import pysam
+import csv
+
+"""
+This script aims to annotate a vcf using:
```
fevac: We are only annotating the info field

```suggestion
This script aims to annotate a vcf info field based on IDs, using:
```

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
+import click
```
fevac, line 5: This is a general script for vcf manipulation that can be used from inputs of any tool. I would therefore rename to "annotate_vcf_by_id.py"

fevac (commented):
I only had time to review the first script! I have a meeting now but I will continue just after. Very nice! Just small comments

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+# parse tsv file
+def parse_tsv(tsv_file: str) -> dict:
+
+    """Parses the tsv file into a dictionary with the id as key and the remaining columns as values in a nested dict"""
+
+    tsv_dict = {}
+    with open(tsv_file, newline='') as f:
+        reader = csv.DictReader(f, delimiter="\t") # convert each tsv row to dict
+
+        id_column = reader.fieldnames[0]
```
beatrizsavinhas: There is no need for these lines to be indented in the `with` statement. As soon as the statement is interrupted, the file is closed. Line 24 already stores the content of the file in `reader`.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+0 and the flag will be added as is to the INFO field based on 0 (false) or 1 (true).
+"""
+
+
+# parse tsv file
+def parse_tsv(tsv_file: str) -> dict:
+
+    """Parses the tsv file into a dictionary with the id as key and the remaining columns as values in a nested dict"""
+
+    tsv_dict = {}
```
beatrizsavinhas: There's no need for this line to be added here far away from its use.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+    with open(tsv_file, newline='') as f:
+        reader = csv.DictReader(f, delimiter="\t") # convert each tsv row to dict
```
beatrizsavinhas: What is the benefit of reading the csv file into a dict? Why not a dataFrame as above?

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+3. a TXT file with the header lines of the tags added, following the VCF specifications (i.e. ##INFO=<ID=ID,Number=number,Type=type,Description="description")
+
+NOTE: the script adapts the annotation style to the Types in the header.
+This means that the ‘Flag’ type indicates that the INFO field does not contain a Value entry, and hence the Number should be
+0 and the flag will be added as is to the INFO field based on 0 (false) or 1 (true).
+"""
+
+
+# parse tsv file
+def parse_tsv(tsv_file: str) -> dict:
```
beatrizsavinhas:
```suggestion
def parse_tsv(tsv_file_path: click.Path) -> dict:
```

`tsv_file` is a very ambiguous name - Is the whole file content included in the string? This would be my assumption when reading this name.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
beatrizsavinhas: Why is this the responsibility of this function? Below you call `merged_linx_files` which, just reading its name you would assume it receives files so this loading should be done there.

beatrizsavinhas (changes requested):
I have really a lot of comments so I suggest we sit and work together on this, otherwise the review process will be too long and frustrating!

bin/combine_linx_tsv.py
kristinebilgrav: There will be duplicates since we are fetching both the 3'prime and 5'prime end and then fetching the svID and later VCF entry for both. If the svID is the same and pointing to the same vcf entry, it will therefore be listed twice.

fevac: aha! yes! then this is good!

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
fevac: I think we could agree on using black for python, as done in other CG tools

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+    sv_file: click.Path,
+    output_file: click.Path) -> None:
+
+    # load tsv into pandas df
+    fusions: pd.DataFrame   = pd.read_csv(fusion_file, sep='\t', dtype=str)
+    breakends: pd.DataFrame = pd.read_csv(breakend_file, sep='\t', dtype=str)
+    svs: pd.DataFrame       = pd.read_csv(sv_file, sep='\t', dtype=str)
+
+    # run main function
+    merged_df: pd.DataFrame = merged_linx_files(fusions, breakends, svs)
```
fevac: It's good practice to name functions starting with verbs to indicate action (merge* get* load* etc)
```suggestion
    merged_df: pd.DataFrame = merge_linx_files(fusions, breakends, svs)
```

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
```diff
@@ -0,0 +1,87 @@
...
+
+    """ Merges input tsv files and returns a dataframe with the relevant columns for annotation of VCF file"""
+
+    # rename columns to desired header names in subsequenct VCF annotation step and change reported column to 0/1
+    fusions.rename(columns ={'name': 'FUSION_NAME', 'reported': 'REPORTED'}, inplace=True)
+    fusions['REPORTED'] = fusions['REPORTED'].replace({'false': 0, 'true': 1}) # to adhere to pysam
+
+    # subset dataframes
+    fusions_subset = fusions[['fivePrimeBreakendId', 'threePrimeBreakendId', 'FUSION_NAME', 'REPORTED']]
+    breakends_subset = breakends[['id', 'svId']]
```
fevac: since you are subsetting the relevant columns to keep at the end I believe you can remove this step here.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+        new_header_lines = hf.read().splitlines()
+
+    for line in new_header_lines:
+        vcf_in.header.add_line(line)
+    merged_header = vcf_in.header
+
+    vcf_out: pysam.VariantFile = pysam.VariantFile( output_file, "w", header=merged_header ) # auto-detects compression from file extension
+
+    for record in vcf_in:
+        if record.id in list(tsv_dict.keys()):
```
fevac:
```suggestion
        if record.id in tsv_dict:
```
why converting into a list? if feels like you would lose the benefits of using dictionaries in terms of speed

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+            annotations = tsv_dict[record.id]
+            record.info.update(annotations)
```
fevac: this can be it's own function (e.g annotate_vcf_entries, update_info_fieldṣ...)

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
fevac: it might has to do with speed, dics are fast when matching based on 'key'. I'm not sure about the speed on pandas but I'd assume it would be slower?

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
fevac: do you need to fix the type from string to what is relevant? is it required by pysam? otherwise it might not be needed-

Regardless, for readability it's nice to do small functions with names that are very self explanatory. 

Also, this only accounts for int and strings not floats. IF this is needed maybe it would be good to include that possibility too.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+        for row in reader:
+            key = row.pop(id_column) # removes id column from dict, returns id (value)
+
+            # convert numeric strings to ints
+            values = { k: int(v) if v.isdigit() else v
+                      for k, v in row.items() }
+
+            # avoid overwriting if id has multiple annotations
+            if key not in tsv_dict:
+                tsv_dict[key] = values
```
fevac: why would it have multiple annotations?

fevac: what is the expectation by pysam if given mulltiple annotations? to have them as lists? or comma separated strings?

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+
+"""
+This script aims to annotate a vcf using:
+1. a VCF file following the VCF specifications, where the ID field will be used to annoate entries
+2. a TSV file where the ids to annotate is always the first column and the remaining are new INFO fields: ex.: ID, TAG1, TAG2, TAG3
+3. a TXT file with the header lines of the tags added, following the VCF specifications (i.e. ##INFO=<ID=ID,Number=number,Type=type,Description="description")
+
+NOTE: the script adapts the annotation style to the Types in the header.
+This means that the ‘Flag’ type indicates that the INFO field does not contain a Value entry, and hence the Number should be
+0 and the flag will be added as is to the INFO field based on 0 (false) or 1 (true).
```
fevac, line 18: is this what you mean?

```suggestion
The script automatically adapts annotation parsing based on the INFO field type defined in the header.
- Fields with Type=Flag are treated as boolean indicators and do not have an associated value in the INFO field.
- For Flag fields should be set to 0 (false) or 1 (true) on the annotation tsv file.
```

fevac: if this is what you mean, it's not fully correct since you are not using the type information given in the header for the value coercion, which would be the ideal way. This could be also done later as part of different PR too.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,103 @@
...
+
+    vcf_in: pysam.VariantFile = pysam.VariantFile(vcf_file, 'rb')
+
+    # add new header lines to original header from header file
+    with open(header_file, 'r') as hf:
+        new_header_lines = hf.read().splitlines()
+
+    for line in new_header_lines:
+        vcf_in.header.add_line(line)
+    merged_header = vcf_in.header
```
fevac: add it in a separate function that can be called here eg. `update_vcf_header`

fevac (commented):
Nice! parse_tsv it's a bit unclear to me the expectation for pysam for types and multi value entries. I assume you added extra steps because you need to coerce types and treat multi value in a special way. but please confirm.

In general it's nice with small python functions that can be easily called and make the code easy to read and follow.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: True!

bin/combine_linx_tsv.py
kristinebilgrav: With the way the next step is currently designed, these columns will either have to be added to the header file and annotated in the VCF - or I could add a part where I ignore columns in the tsv if they are not in the header file or similar.

modules/local/linx_annotation/linx_tsv/scripts/combine_linx_tsv.py
kristinebilgrav: Moved to the function

bin/annotate_vcf_by_id.py
kristinebilgrav: Do you think it should also be lifted out from `linx_annotation` and be its own module?

bin/annotate_vcf_by_id.py
kristinebilgrav: I think so. Pysam will automatically add information in the correct manner based on the header lines. So if the annotation type is a Flag it expects the number in the header line to be 0 (standard VCF format guidelines). And then when you give pysam what you want to annotate with it expects something like FLAG_TAG = 0/False. 

But not sure what you mean here with using the header information and a separate PR?

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: Exactly, this is faster than using pandas.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: Yes, pysam want the 0 and 1 to be int specifically. 
But true for the floats! And I will make it its own small function

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: Sometimes there are multiple fusions connected to the same VCF entry. And I assume we want all names to be added. Pysam then wants a tuple like so; KEY : (NAME1, NAME2). So I first make a list if there are more than one name and then after make it a tuple.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: Hm, not sure, might have experienced an error or so, but works - so I will add this!

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: I think when adding click.Path it becomes clear that its a path, but will change this parameter.

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: It describes what the function does. I can try to make it more descriptive of the function

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
kristinebilgrav: When I unindent this it throws an error `ValueError: I/O operation on closed file.` So I do think the file needs to be open 🤔

bin/combine_linx_tsv.py
fellen31, line 1: Didn't see a linked issue so not sure if the approach has already been discussed. But could this script be replaced by [csvtk](https://github.com/shenwei356/csvtk), which already has a lot of nf-core modules?

bin/annotate_vcf_by_id.py
fellen31, line 1: Not sure I have the full context here, but would `bcftools annotate` be an alternative? Do you have to match on ID, or could you use `CHROM,POS,REF,ALT`?

bin/combine_linx_tsv.py
kristinebilgrav: Sorry, missed adding the issue! Its here https://github.com/Clinical-Genomics/MTP-oncoflow/issues/67. 

I had not seen this tool, but it looks very promising! I think this could replace it by using `join`, `rename`, `replace` and `cut/split`. That would mean writing a few nf-core modules though, as there are only modules for `join`  and `split`. 
What do you think @fevac? I think that we should aim to get this done, but then in the future replace it with this tool perhaps?

bin/annotate_vcf_by_id.py
kristinebilgrav: I did explore that, but I would have to run `bcftools` twice; once to extract entries with ID and then again to annotate with the retrieved `CHROM,POS,REF,ALT`.
We only have the ID from the TSVs

fellen31: Okay, personally I would still prefer using standard bioinformatics tools for the task even if it means running it twice.

bin/combine_linx_tsv.py
fellen31: I wrote the mutate2 module a few months ago, it's pretty quick since you can reuse a lot of from the other modules. Bit inconvenient that you would need 4 different modules though.

bin/annotate_vcf_by_id.py
kristinebilgrav: I understand that, and this was my first suggested approach, however we collectively decided on a script in a refinement session earlier. Happy to revisit though!

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,191 @@
...
+2. a TSV file where the ids to annotate is always the first column and the remaining are new INFO fields: ex.: ID, TAG1, TAG2, TAG3
+3. a TXT file with the header lines of the tags added, following the VCF specifications (i.e. ##INFO=<ID=ID,Number=number,Type=type,Description="description")
+
+NOTE: the script adapts the annotation style to the Types in the header.
+This means that the ‘Flag’ type indicates that the INFO field does not contain a Value entry, and hence the Number should be
+0 and the flag will be added as is to the INFO field based on 0 (false) or 1 (true).
+"""
+
+
+def define_data_types(row: dict) -> dict:
```
emmadizdarevic: Would it be possible to add "more detailed" type hints, e.g. `dict[str, dict[str, int]]` or `tuple[str, int]`... ?

modules/local/linx_annotation/annotate_vcf_by_id/scripts/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,191 @@
...
+                continue
+
+            # current entry for column
+            current_entry = tsv_annotations_dict[vcf_id][annotation_tag]
+            if isinstance(current_entry, list):
+                current_entry.append(annotation_value)
+            else:
+                tsv_annotations_dict[vcf_id][annotation_tag] = [current_entry, annotation_value]
+
+    return tsv_annotations_dict
```
emmadizdarevic: Speaking of atomic functions, this one could probably be split into 2-3 smaller ones. I don't have a clear solution but maybe the second part (after else) can be its own function that just handles the merging of new annotation values.

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,150 @@
...
+)
+def combine_linx_files(
+    fusion_file: click.Path,
+    breakend_file: click.Path,
+    sv_file: click.Path,
+    output_file: click.STRING,
+) -> None:
+
+    # run main function
+    merged_df: pd.DataFrame = merge_linx_files(fusion_file, breakend_file, sv_file)
```
beatrizsavinhas: Follow Single responsibility principle. The main responsibility of `merge_linx_files` is to produce a merged dataframe but it is actually and implicitly also responsible for opening and loading files. This would remove that hidden responsibility and clarify that `merge_linx_dataframes` simply receives 3 dataframes, manipulates them and returns a joint dataframe which is much closer to what its name says.
 
```suggestion
	# Load input files into pandas df
    fusions: pd.DataFrame = pd.read_csv(fusion_file, sep="\t", dtype=str)
    breakends: pd.DataFrame = pd.read_csv(breakend_file, sep="\t", dtype=str)
    svs: pd.DataFrame = pd.read_csv(sv_file, sep="\t", dtype=str)

    # Merge all dataframes
    merged_df: pd.DataFrame = merge_linx_dataframes(fusions, breakends, svs)
```

bin/combine_linx_tsv.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def merge_dataframes(
+    df1: pd.DataFrame, df2: pd.DataFrame, left_on: str, right_on: str, how: str
+) -> pd.DataFrame:
+    """Merges two dataframes based on specified columns and merge method"""
+
+    merged_df = df1.merge(df2, left_on=left_on, right_on=right_on, how=how)
+
+    return merged_df
```
beatrizsavinhas: Don't reinventing the wheel - The native operators already fulfil this logic need so there's no need to add another layer of abstraction.
I suggest removing these functions.

assets/sv_annotation_header.txt
beatrizsavinhas, line 1: I'm thinking if it would be good that the logic itself can generate this header based on the fields that the code adds which would reduce the amount of maintenance and manual work required to keep this updated and also the risk of errors with the header accidentally not matching what is changed in the VCF file.
I wouldn't say something like this should be added now, but perhaps something to consider in the future! 🤔

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+    output_index: pysam.TabixFile = pysam.tabix_index(
+        output_file, preset="vcf", force=True
+    )
+
+    return output_file, output_index
+
+
+@click.command()
+
+# get input files
```
beatrizsavinhas:
```suggestion
```
/nit avoid comments when the code is explanatory.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+    "-h",
+    "--header_file",
+    type=click.Path(exists=True),
+    help="A txt file with new header lines for the annotations",
+    required=True,
+)
+@click.option(
+    "-o",
+    "--output_file",
+    type=click.Path(),  # change to string
```
beatrizsavinhas: Seems like this was missed?

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+    tsv_annotations_dict = {}
+    with open(tsv_file_path, newline="") as f:
+        # converts each tsv row to dict
+        tsv_reader: csv.DictReader = csv.DictReader(f, delimiter="\t")
+
+        vcf_id_column_header: str = tsv_reader.fieldnames[0]
+
+        for tsv_row_dict in tsv_reader:
+            # removes id column from dict, returns id
+            vcf_id: str = tsv_row_dict.pop(vcf_id_column_header) # maybe split up
```
beatrizsavinhas: More notes here that seem to have been missed.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+    for key, value in tsv_row_dict.items():
+        try:
+            tsv_row_dict[key] = int(value)
+        except ValueError:
+            try:
+                tsv_row_dict[key] = float(value)
+            except ValueError:
+                tsv_row_dict[key] = str(value)
+
+    return tsv_row_dict
```
beatrizsavinhas: After a quick google search it seems that `ast.literal_eval` already performs this logic.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+def get_dict_from_tsv(tsv_file_path: click.Path) -> dict:
+    """Parses the tsv file into a dictionary (id as key and the remaining columns as values in a nested dict),
+    to be compatible with annotation with pysam"""
+
+    tsv_annotations_dict = {}
+    with open(tsv_file_path, newline="") as f:
+        # converts each tsv row to dict
+        tsv_reader: csv.DictReader = csv.DictReader(f, delimiter="\t")
+
+        vcf_id_column_header: str = tsv_reader.fieldnames[0]
```
beatrizsavinhas: There is no need and in fact it's much safer to not include more code in the `with` block. As soon as the block ends, the file is closed. The more actions are performed with an open file, the higher the changes of corrupting the file if something goes wrong unexpectedly.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+    header_file: click.Path,
+    tsv_file: click.Path,
+    output_file: click.Path,
+) -> None:
+
+    # parse tsv file
+    tsv_dict: dict = get_dict_from_tsv(tsv_file_path=tsv_file)
+
+    # annotate vcf
+    annotate_vcf(vcf_file, header_file, tsv_dict, output_file)
```
beatrizsavinhas: This function seems to be doing a lot of things at the same time.

Just reading this code, I'd expect some logic like this for the `main` function:
1. Get dictionary from file.
2. Manipulate dictionary to have the desired structure.
3. Print dictionary to VCF file.

This is how I would suggest refactoring the code to try to simplify the logic and responsibilities of each function.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,251 @@
...
+
+    vcf_in.close()
+    vcf_out.close()
+
+    # write index file for annotated vcf
+    output_index: pysam.TabixFile = pysam.tabix_index(
+        output_file, preset="vcf", force=True
+    )
+
+    return output_file, output_index
```
beatrizsavinhas: What are these returns for? The function is called without any capturing any return in line 246.

beatrizsavinhas (changes requested):
I again ran out of time so I couldn't finish the review of `annotate_vcf_by_id.py` but tried to leave some overarching comments below. Please also check the errors that the IDE signals.

bin/combine_linx_tsv.py
beatrizsavinhas, line 1: The code in this file is looking much cleaner! 🧹 ✨

bin/annotate_vcf_by_id.py
kristinebilgrav: Added, but this module interprets strings as variable names and will fail for strings.

assets/sv_annotation_header.txt
kristinebilgrav: Thats a good idea! However I think that this can be a point to add in the future.

bin/annotate_vcf_by_id.py
kristinebilgrav: This part loops over the rows of the file, so I do need to add it to the dictionary here. I could move the dictionary modifications to a different place, but that would result in looping through the information twice, which seems a little unecessary?

bin/annotate_vcf_by_id.py
kristinebilgrav: Will remove

bin/annotate_vcf_by_id.py
kristinebilgrav: As mentioned briefly here: [comment][https://github.com/Clinical-Genomics/oncorefiner/pull/114/changes#r3459048521), I would still need to do `add_entry_to_annotation_dict` to accomodate for multiple entries to the same ID, but could lift out `convert_data_type`. However I think that its a bit unnecessary to loop over the dictionary to do only this?

bin/combine_linx_tsv.py
kristinebilgrav: I agree haha. Didnt know where your limit of functions went 👍

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def annotate_vcf(
+    vcf_file: click.Path,
+    header_file: click.Path,
+    tsv_dict: dict[str, dict[str, str | int | float | tuple]],
+    output_file: click.Path,
+) :
+    """Annotate VCF file using the tsv_dict and header_file, and save the annotated VCF to output_file.
+    Returns the path to the annotated VCF and its index file"""
```
beatrizsavinhas: This information is incorrect.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+        record: pysam.VariantRecord = update_vcf_info_field(record, tsv_dict)
+        vcf_out.write(record)
+
+    vcf_in.close()
+    vcf_out.close()
+
+    # write index file for annotated vcf
+    output_index: pysam.TabixFile = pysam.tabix_index(
+        output_file, preset="vcf", force=True
+    )
```
beatrizsavinhas: This is yet another hidden responsibility. Suggest removing this from this function and writing it separately, explicitly.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+                add_entry_to_annotation_dict(
+                    vcf_id, vcf_id_annotations, tsv_annotations_dict
+                )
+            )
+
+    # convert list to tuple if multiple values for same field
+    tsv_annotations_dict: dict[str, dict[str, str | int | float | tuple]] = (
+        convert_lists_to_tuples(tsv_annotations_dict)
+    )
+    return tsv_annotations_dict
```
beatrizsavinhas: A suggestion to address my previous comments:
```suggestion
def get_dict_from_tsv(tsv_file_path: click.Path) -> dict:
    """Parses a tsv file into a dictionary with format: {id: [{field: value, ...}, {field: value, ...}, ...]},
    compatible with annotation with pysam"""
    
    with open(tsv_file_path, newline="") as f:
        raw_annotation_dict_list: list[dict] = [row for row in csv.DictReader(f, delimiter="\t")]

    annotation_dict_list: list[dict] = [format_annotation_dict(raw_annotation_dict) for raw_annotation_dict in raw_annotation_dict_list]

    return merge_dicts(*annotation_dict_list)
```

Given:

```python
def string_to_object(string: str) -> str | int | float:
    """Convert a string to the appropriate variable type (int, float, bool, etc.). If the string cannot be converted, return the original string."""
    try:
        return ast.literal_eval(string)
    except:
        return string
        
def format_annotation_dict(raw_annotation_dict: dict) -> dict:
    vcf_id_key: str = next(iter(raw_annotation_dict))
    vcf_id: str = raw_annotation_dict.pop(vcf_id_key)
    info_fields_dict: dict = {key: string_to_object(value) for key, value in raw_annotation_dict.items()}
    return {vcf_id: info_fields_dict}

def merge_dicts(*dicts):
    merged_dict = {}
    for dict in dicts:
        for key in dict:
            try:
                merged_dict[key].append(dict[key])
            except KeyError:
                merged_dict[key] = [dict[key]]
    return merged_dict
````

> [!NOTE]
> This produces a dictionary with a format like
> ```
> {'vcf_id_1': [{'FUSION_NAME': 'name', 'REPORTED': 0}, {'FUSION_NAME': 'other_fusion_name', 'REPORTED': 0}],
> 'vcf_id_2': [{'FUSION_NAME': 'name', 'REPORTED': 1}]}
> ```
> that perhaps is slightly different than this implementation so testing is needed to make sure it is accepted by `pysam`.

File: bin/annotate_vcf_by_id.py [hunk 57]
```diff
@@ -0,0 +1,246 @@
...
+    )
+
+    # pysam auto-detects compression from file extension
+    vcf_out: pysam.VariantFile = pysam.VariantFile(
+        output_file, "w", header=merged_header
+    )
+
+    # need to loop through records to get entries with correct ID
+    for record in vcf_in:
+        record: pysam.VariantRecord = update_vcf_info_field(record, tsv_dict)
```
beatrizsavinhas:
```suggestion
        updated_record: pysam.VariantRecord = record.info.update(tsv_dict[record.id])
```

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+
+def update_vcf_info_field(
+    record: pysam.VariantRecord, tsv_dict: dict[str, dict[str, str | int | float | tuple]]
+) -> pysam.VariantRecord:
+    """Updates the INFO field of a VCF record with annotations from the tsv_dict"""
+
+    if record.id in tsv_dict:
+        annotations: dict[str, str | int | float | tuple] = tsv_dict[record.id]
+        record.info.update(annotations)
+    return record
```
beatrizsavinhas: Don't reinvent the wheel; native functions take precedence. See https://github.com/Clinical-Genomics/oncorefiner/pull/114/changes#r3476011687.
```suggestion
```

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+        output_file, "w", header=merged_header
+    )
+
+    # need to loop through records to get entries with correct ID
+    for record in vcf_in:
+        record: pysam.VariantRecord = update_vcf_info_field(record, tsv_dict)
+        vcf_out.write(record)
+
+    vcf_in.close()
+    vcf_out.close()
```
beatrizsavinhas: Is it possible to avoid having both files open simultaneously?
If so, list comprehension would look clean here.
```suggestion
	updated_variants: list[pysam.VariantRecord] = [record.info.update(tsv_dict[record.id]) for record in vcf_in]
    vcf_in.close()
    
    # Open vcf_out file
    # Update vcf_out file header
    # Write updated_variants to vcf_out file
    
    vcf_out.close()
```

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def update_vcf_header_with_new_lines(
+    vcf_in: pysam.VariantFile, header_file_path: click.Path
+) -> pysam.VariantHeader:
+    """ Adds new header lines to original header from header file, and returns the merged header"""
+
+    # add new header lines to original header from header file
+    with open(header_file_path, "r") as hf:
+        new_header_lines = hf.read().splitlines()
```
beatrizsavinhas:
```suggestion
    with open(header_file_path, "r") as header_file:
        new_header_lines = header_file.readlines()
```

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,246 @@
...
+
+    for vcf_id, vcf_id_annotations in tsv_annotations_dict.items():
+        for annotation_tag, annotation_value in vcf_id_annotations.items():
+            if isinstance(annotation_value, list):
+                vcf_id_annotations[annotation_tag] = tuple(annotation_value)
+
+    return tsv_annotations_dict
+
+
+def update_vcf_header_with_new_lines(
```
beatrizsavinhas:
```suggestion
def get_updated_vcf_header(
```

beatrizsavinhas (approved):
In the interest of time, I think we should move on with this PR.
There are several aspects that I think should be improved so I left suggestions below. Feel free to ask for clarifications since I did not have time to write detailed descriptions.

bin/annotate_vcf_by_id.py
beatrizsavinhas: P.S.: We can modify `merge_dicts` to collapse `'FUSION_NAME'` but I was unsure of how it should be done for `'REPORTED'`.

bin/annotate_vcf_by_id.py
kristinebilgrav: Moved to the main part

bin/annotate_vcf_by_id.py
kristinebilgrav: It would probably be very clean, but we would also have to store the whole vcf in a list and I feel like that would be an unnecessary step..

bin/annotate_vcf_by_id.py
kristinebilgrav: Talked, and implementing some of this

bin/combine_linx_tsv.py
kristinebilgrav: Not adding columns for now. Will keep a note in the file for now

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,166 @@
...
+        new_header_lines = header_file.readlines()
+
+    for line in new_header_lines:
+        vcf_in.header.add_line(line)
+    merged_header = vcf_in.header
+
+    # create a dictionary to store the type of each header ID
+    header_id_type = {}
+    for name, meta in merged_header.info.items():
+        header_id_type[name] = meta.type
```
beatrizsavinhas, line 37: As I mentioned yesterday, this dictionary only needs to contain the fields from the header file since they are the ones that will be present in the tsv file. 
This should be moved to its own function - there's no indication of this responsibility in the name or docstring of this function anyway.

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,166 @@
...
+        for field, values in fields.items():
+            if len(values) == 1:
+                continue
+            else:
+                if header_id_type.get(field) == "Flag":
+                    grouped[key][field] = max(values)
+                else:
+                    grouped[key][field] = tuple(values)
+
+    return grouped
```
beatrizsavinhas: I don't have a header file example so I can't test it locally, but see if this works as intended!

```suggestion
def merge_dict_by_field(dicts: list[dict], field_type: dict[str, str]):
    """Merges a list of dictionaries into a single dictionary, grouping values by key, based on field type. If a field is of type "Flag", the maximum
    value is kept; otherwise, all values are stored in a tuple.
    
    Keyword arguments:
    dicts -- list of dictionaries to merge i.e. [{vcf_id: {field: value, ...}}, {vcf_id: {field: value, ...}}, ...]  
    field_type -- dictionary mapping field names to their types i.e. {field_name: "Flag" | "Other", ...}
    """
    
    merged_dict = {}

    for dict in dicts:

        vcf_id: str = next(iter(dict))
        
        if vcf_id not in merged_dict:
            merged_dict[vcf_id] = dict[vcf_id]
        else:
            for field, value in dict[vcf_id].items():
                if field_type[field] == "Flag":
                    merged_dict[vcf_id][field] = max(merged_dict[vcf_id][field], value)
                else:
                    # Append to existing list or create list with existing value and new value
                    if isinstance(merged_dict[vcf_id][field], list):
                        merged_dict[vcf_id][field].append(value)
                    else:
                        merged_dict[vcf_id][field] = [merged_dict[vcf_id][field], value]

    return merged_dict
```

This is not great as the if else tree if extensive. But it already avoids iterating through the dictionary twice.

bin/annotate_vcf_by_id.py
kristinebilgrav: I could only use the pysam functions on an actual VCF header, thus its done after the addition of the new lines to the new header. And since it was such a small function, I though it was OK to keep here. I could do a manual processing of the file to a dictionary in a different function if you prefer that?

bin/annotate_vcf_by_id.py
kristinebilgrav: Seems to be working, thanks!

bin/annotate_vcf_by_id.py
beatrizsavinhas: I think it would be cleaner to separate responsibilities and not have an ambiguous `prepare_header` function. For that you would probably have to separate the loading of the header file into `pysam.VariantHeader` too, which would require spending some more time on this so it's up to you.
As mentioned below, I think we can move on and open an issue to refactor these scripts and/or even use existing tools like Felix suggested above!

kristinebilgrav: Sounds good, will open an issue!

PR #115, opened 2026-05-29
feat: Annotate sv linx

kristinebilgrav:
WIP 
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/67

### Added

- Script to merge LINX tsv files https://github.com/Clinical-Genomics/oncorefiner/pull/114
- Script to annotate SV vcf with information from LINX files https://github.com/Clinical-Genomics/oncorefiner/pull/114
- Parameters `linx_fusion_tsv , linx_breakend_tsv  and linx_sv_tsv ` for LINX files from OA and test files https://github.com/Clinical-Genomics/oncorefiner/pull/118
- The file `sv_annotation_header.txt` https://github.com/Clinical-Genomics/oncorefiner/pull/118
- Added local modules `COMBINE_LINX_TSV` and `ANNOTATE_VCF_BY_ID` https://github.com/Clinical-Genomics/oncorefiner/pull/140
- Added the subworkflow `ANNOTATE_LINX` https://github.com/Clinical-Genomics/oncorefiner/pull/143

### Changed

- Added the subworkflow `ANNOTATE_LINX` to `PROCESS_SVS` https://github.com/Clinical-Genomics/oncorefiner/pull/143
- Updated tests and snapshots https://github.com/Clinical-Genomics/oncorefiner/pull/143
 
### Fixed

-

### Removed

-


bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,191 @@
...
+
+import click
+import pysam
+import csv
+import ast
+
+"""
+This script aims to annotate the info field of a vcf based on the ID field, using:
+1. a VCF file following the VCF specifications, where the ID field will be used to annotate entries
+2. a TSV file where the ids to annotate is always the first column and the remaining are new INFO fields: ex.: ID, TAG1, TAG2, TAG3
```
mathiasbio, line 11: Have we talked with the scout-devs about the formats of these fields, and how they could be made use of in Scout for displaying information and maybe for filtering? : )

bin/annotate_vcf_by_id.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+
+import click
+import pysam
+import csv
+import ast
+
+"""
+This script aims to annotate the info field of a vcf based on the ID field, using:
+1. a VCF file following the VCF specifications, where the ID field will be used to annotate entries
```
mathiasbio, line 10: I have another question now that I have looked into the format of the Purple SV VCF a bit more. It seems that all variants (?), that are not SGL (single break ends) have 2 entries in the VCF per variant. 

From what I can tell, if we follow the trace through these files to the fusion annotation, we will have:

1. a fusion in the fusion.tsv which has 2 break-ends associated
2. via the break-end.tsv file we can see the SV-ID
3. in the SV.tsv we will see that the SV-ID is associated with a VCF ID. 

But! This VCF ID is only the first of the entries of the variant. 

So for instance for in the variant discussed here: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/90 

There are 2 breakends associated to the deletion in the VCF

chrX	1212762	46042	A	ATTTCGAAGGGG[chrX:1535844[
chrX	1535844	46104	G	]chrX:1212762]TTTCGAAGGGGG	 

But only 46042 is linked as the SV ID. 

Do you know if we have ever discussed the possibility of merging these two breakpoint-level entries in the purple VCF to one entry per variant instead?

mathiasbio: Or how we otherwise plan to handle these types of variants downstream? I feel like we're not used to this format in Scout when it comes to deletions and duplications 🤔 But only for translocations

mathiasbio: I'm thinking along the lines of merging the two entries due to starting to look into this issue: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/94

There it was proposed to change all DEL svtypes to BND as a solution to VEP not being able to annotate these variants. Which I suppose works because VEP then accepts the format of having this type of info "ATTTCGAAGGGG[chrX:1535844[" in the ALT column of the VCF. 

But I wonder if it wouldn't be better to merge the two entries into one, and have a more standard DEL representation, which then VEP should be able to handle as well

bin/annotate_vcf_by_id.py
kristinebilgrav: Hm, no not really! First they will be used to make a scoring. So the field we want to add is `REPORTED` and then we use GENMOD to score that, and this will be used as a ranking in scout. Do you think they could cause problems or could we ask them at a later stage when we perhaps have more to add?

bin/annotate_vcf_by_id.py
kristinebilgrav: I noticed this too! And we did talk briefly about them during refinement, but only in regards to the annotation. So we did not discuss to merge. 
I feel like I have seen other VCFs have two representations of the deletion, one from each side, but perhaps I remember incorrectly? 
Regardless, I agree that I would be better to somehow make the calls more compatible with standard DEL representations (ref https://github.com/Clinical-Genomics/MTP-oncoflow/issues/94), I even think that an issue was opened here: https://github.com/hartwigmedical/hmftools/issues/820 to already make the tool do it the standard way, so we dont have to make more work-arounds and that hopefully issue #94  could have a temporary fix, but I am unsure if they will prioritize it. 
Maybe we could set up at refinement about it when everyone is back from vacation?

mathiasbio: Yeah I think this needs a discussion to agree on a solution, maybe the others have more context about this, and I might have missed something. I'll try to explore the issue a bit more and write some suggested solution in the meantime. 

Regardless of if we decide to merge the 2 entires into 1 per variant (except for translocations), the solution in this PR shouldn't be affected. Though if we don't merge them, I feel like it caaan be a bit confusing that we have in essence the same variant from two perspectives, but only one is annotated with this linx information. Maybe we can note this down somewhere just so we don't forget 🤔

bin/annotate_vcf_by_id.py
mathiasbio: Probably we can look into it later 🙏

bin/annotate_vcf_by_id.py
kristinebilgrav: True! I agree. I added some documentation about it in the [output.md](https://github.com/Clinical-Genomics/oncorefiner/pull/115/changes#diff-be641ba3494830f94b1cddc711570ff7c4ebe20bbb311dec475d8da0ffd82aa0) file, what do you think?

kristinebilgrav: Merging in order to move forward with the next part dependent on this (https://github.com/Clinical-Genomics/MTP-oncoflow/issues/329), but let me know what you think and we can fix afterhand!

mathiasbio: That's perfect thank you! ❤️

PR #116, opened 2026-06-01
fix: Update `GENERATE_CYTOSURE_FILES` subworkflow test snapshot

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/71.

### Fixed

- Updated `GENERATE_CYTOSURE_FILES` subworkflow test snapshot. The changes introduced in https://github.com/Clinical-Genomics/oncorefiner/pull/110 result in the generation of a modified output file for the subworkflow. However, for some unknown reason, the test for this subworkflow was either not triggered or incorrectly set as passed in the previous PR but caused now fails in more recent PRs.


kristinebilgrav (approved):
👍

PR #117, opened 2026-06-01
feat: Add stub tests

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/72.

### Added

- Stub tests for subworkflows:
	- `GENERATE_CYTOSURE_FILES`
	- `PREPARE_REFERENCES`
	- `PROCESS_SNVS`
	- `PROCESS_SVS`

- Stub tests for both pipeline tests:
	- `Tumor Normal`
	- `Tumor only` 

### Changed

- Fixed name of process SVs to match other tests.


subworkflows/local/process_svs/tests/main.nf.test
```diff
@@ -1,6 +1,6 @@
 nextflow_workflow {
 
-    name "Test subworkflow Process Structural Variants (PROCESS_SVS)"
```
beatrizsavinhas, line 3: I noticed that the test name was inconsistent with the rest of tests. This change is quite out of scope but I thought it would be inconsequential enough to add here. Will gladly remove it if you don't agree!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -49,4 +49,50 @@ nextflow_workflow {
...
+            def output_files = getAllFilesFromDir(params.outdir)
+
+            assertAll(
+                { assert workflow.success },
+                { assert snapshot(
+                    // All directories and files in the output directory
+                    output_directories_and_files,
+                    // All files in the output directory with their md5 checksum
+                    output_files.collect { file -> [ file.getName(), path(file.toString()).md5 ] }
+                ).match() }
```
fellen31, line 90: Perhaps? Or do you think this is a separate issue? 

```suggestion
                { assert snapshot(workflow.out).match() }
```

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas, line 1: Clarification: All tests follow the same structure the non-stub tests, except only the output file names and directories are collected!

tests/tumor_normal.nf.test
fellen31, line 1: Could perhaps consider putting the new stub test in a separate file. Not sure what you think is best. If you add a test where you skip genmod score, I think that could be good to separate.

tests/tumor_normal.nf.test
```diff
@@ -39,4 +39,33 @@ nextflow_pipeline {
...
+            def stable_path = getAllFilesFromDir(params.outdir, ignoreFile: 'tests/.nftignore')
+            assertAll(
+                { assert workflow.success},
+                { assert snapshot(
+                    // pipeline versions.yml file for multiqc from which Nextflow version is removed because we test pipelines on multiple Nextflow versions
+                    removeNextflowVersion("$outputDir/pipeline_info/oncorefiner_software_mqc_versions.yml"),
+                    // All stable path name, with a relative path
+                    stable_name,
+                    // All files with stable contents
+                    stable_path
```
beatrizsavinhas: For this and `tests/tumor_only.nf.test`, all files can be collected since they should be empty files and therefore have stable content.

tests/tumor_only.nf.test
```diff
@@ -40,4 +40,33 @@ nextflow_pipeline {
...
+            def stable_path = getAllFilesFromDir(params.outdir, ignoreFile: 'tests/.nftignore')
+            assertAll(
+                { assert workflow.success},
+                { assert snapshot(
+                    // pipeline versions.yml file for multiqc from which Nextflow version is removed because we test pipelines on multiple Nextflow versions
+                    removeNextflowVersion("$outputDir/pipeline_info/oncorefiner_software_mqc_versions.yml"),
+                    // All stable path name, with a relative path
+                    stable_name,
+                    // All files with stable contents
+                    stable_path
```
fellen31: I don't think you should need to check the md5sums of the stub files?

fellen31 (commented):
🌟

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: Hahaha! I thought exactly the same! But for now the subworkflow doesn't have any outputs so they can't be checked. This should be fixed in https://github.com/Clinical-Genomics/CancerBioInfo/issues/112!

tests/tumor_only.nf.test
beatrizsavinhas: I thought we could make sure they are empty? But yeah, we don't know what each tool will produce in stub, so it probably makes sense to not check. 🤔

tests/tumor_normal.nf.test
beatrizsavinhas: Humm I think these general stubtests are OK to include in the same file. When we start adding tests for all different input conditions and the file gets too long, I agree we should consider moving the stub tests to a separate file!

subworkflows/local/generate_cytosure_files/tests/main.nf.test
fellen31: Aha, okay.

tests/tumor_only.nf.test
fellen31: Yeah, some modules produce empty files, others do not, so I actually think it's a bit more confusing therefore.

beatrizsavinhas: Will remove then!

beatrizsavinhas: Done now!

kristinebilgrav (approved):
LGTM!

PR #119, opened 2026-06-02
feat: add gens files

mathiasbio:
Closes: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/95

### Added

- Add `cobalt_ratio_pcf_tumor` and `cobalt_ratio_pcf_normal` to params https://github.com/Clinical-Genomics/oncorefiner/pull/112
- Add `amber_baf_tsv_gz` to params https://github.com/Clinical-Genomics/oncorefiner/pull/112
- Add param `sampletype` https://github.com/Clinical-Genomics/oncorefiner/pull/112
- Add process amber baf gens module https://github.com/Clinical-Genomics/oncorefiner/pull/120
- Add process cobolt gens module https://github.com/Clinical-Genomics/oncorefiner/pull/122
- Add subworkflow *TBD* for to PROCESS_CNVS

### Changed

- Added file paths for testdataset for above mentioned params  https://github.com/Clinical-Genomics/oncorefiner/pull/112
- Added necessary files to testdatasets
- Updated pipeline test snapshot

### Fixed

-

### Removed

-


nextflow.config
```diff
@@ -34,6 +34,12 @@ params {
     cadd_resources              = null
     cadd_prescored_indels       = null
 
+    // GENS
+    amber_baf_tsv_gz            = null
+    cobalt_ratio_pcf_normal     = null
+    cobalt_ratio_pcf_tumor      = null
+    analysis_type               = null
```
beatrizsavinhas: See previous discussion on this https://github.com/Clinical-Genomics/oncorefiner/pull/158#discussion_r3828507137.

kristinebilgrav: Added now! It looks at the PCF files (which is needed for `PREPARE_COBALT_FOR_GENS`, whilst this value is for `PREPARE_AMBER_FOR_GENS`), but seemed to be more logical than to look at the T/N BAM files. Also after deciding (in the same issue as above) that we want to always run both processes, this seems ok? Let me know if you know other files/values that are better to look at.

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python3
+
+import click
+import pandas as pd
+
+### tumor and normal baf take tumor/normal parameter to know which file to check.
```
fevac, line 6: I don't understand this comment, please rewrite and clarify

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
...
+
+__version__ = "1.0"
```
fevac, line 8: is the plan to keep updating this version or will be falling out of sync? I don't think this is needed since this is an script within a versioned pipeline and not a standalone one

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
...
+DEFAULT_LEVELS = {
+    "o": 135,
+    "a": 30,
+    "b": 8,
+    "c": 3,
+    "d": 1,
+}
+
+
+def chrom_to_output_name(chromosome: str, level: str) -> str:
```
fevac, line 19: missing description

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
...
+    chrom = str(chromosome).replace("chr", "")
+    return f"{level}_{chrom}"
+
+
+def write_baf_file(
```
fevac, line 24: this function is not only writing the baf file but also doing de conversion. Ideally you would convert/parse in a separate function

fevac, line 24:
```suggestion
def generate_baf_file_for_gens(
```

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
...
+                pos = int(row["position"])
+
+                rows.append(
+                    [
+                        chrom_to_output_name(chrom, level),
+                        pos,
+                        pos + 1,
+                        f"{float(row[baf_column]):.4f}",
+                    ]
+                )
```
fevac, line 46: Move this into it's own function (called something like "create_baf_bed_rows_by_zoom_level") or at the very list add a comment to explain what is going on here and the strategy used to generate the bed file

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,120 @@
...
+    chrom = str(chromosome).replace("chr", "")
+    return f"{level}_{chrom}"
+
+
+def write_baf_file(
+    df: pd.DataFrame,
+    baf_column: str,
+    output_path: str,
+    levels: dict,
+) -> None:
```
fevac, line 29: add description so people can easily know what happens here

bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+
+import math
+import click
+import pandas as pd
+
+__version__ = "1.0"
```
fevac, line 7: similar here

bin/prepare_amber_for_gens.py
fevac: Just to be clear: If we keep it this, it MUST be updated every time there is a change in the script

bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def chrom_to_output_name(chromosome: str, level: str) -> str:
+    chrom = str(chromosome).replace("chr", "")
+    return f"{level}_{chrom}"
+
+
+def segment_to_points(start: int, end: int, spacing: int) -> list[int]:
+    """
+    Split a segment into chunks of approximately `spacing` bp
```
fevac, line 25: `approximately`? I guess you know exactly how it's done. Please be as clear as possible and precise.

fevac: It's also confusing how the function is named and the description given. The function mentions `points` and the description `chunks` and It's difficult to understand what's the point of all this.

If I understand correctly, what you are really doing here is sampling segments by zoom levels and taking their midpoint, aren't you?

bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,125 @@
...
+
+    return points
+
+
+def write_segment_zoom_file(
+    df: pd.DataFrame,
+    output_path: str,
+    levels: dict[str, int],
+) -> None:
+    rows = []
```
fevac, line 49: add description

bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,125 @@
...
+    for i in range(n_points):
+        chunk_start = start + int(i * length / n_points)
+        chunk_end = start + int((i + 1) * length / n_points) - 1
+        midpoint = (chunk_start + chunk_end) // 2
+        points.append(midpoint)
+
+    return points
+
+
+def write_segment_zoom_file(
```
fevac, line 44: this is not only writing but also doing changes in the content. 
also, what segments are these? Log2ratios? Could this be rename to `generate_log2ratios_for_gens` or is that incorrect?

fevac (commented):
Hej! The python scripts might need a bit of love. I've comment on these now. The descriptions missing should be included in this PR. Since we are a bit in a hurry to get this tested, for any other change you can open another issue to do it once this is feature merged. If the latter, please link all the comments not addressed here in the new issue(s)

docs/output.md
```diff
@@ -73,10 +73,16 @@ This process produce an interactive CNV report.
 
 - `cnv`
   - `<meta.id>_cnv_report.html`: an interactive CNV report in html format.
+  - `<meta.id>_CNV_BAF_GENS.<tumor or normal>.baf.zoom.tsv.gz`: a gzipped tsv containing b-allele frequency (BAF) for visualization using GENS, produced from Oncoanalysers AMBER files.
+  - `<meta.id>_CNV_BAF_GENS.<tumor or normal>.baf.zoom.tsv.gz.tbi`: index file for the gzipped tsv with BAF.
```
fevac, line 77:
```suggestion
  - `<meta.id>_CNV_BAF_GENS.<tumor or normal>.baf.zoom.bed.gz`: a gzipped tsv containing b-allele frequency (BAF) for visualization using GENS, produced from Oncoanalysers AMBER files.
  - `<meta.id>_CNV_BAF_GENS.<tumor or normal>.baf.zoom.bed.gz.tbi`: index file for the gzipped tsv with BAF.
```
Aren't these bed files?

docs/output.md
```diff
@@ -73,10 +73,16 @@ This process produce an interactive CNV report.
...
+  - `<meta.id>_<meta.sample_type>_CNV_COV_GENS.bed.gz`: a gzipped bed file with genomic coverage levels for visualization using GENS, produced from Oncoanalysers COBALT files.
```
fevac, line 78: it would be nice to be consistent with the naming, let's keep a similar format for both gens files. Now you have:
<meta.id>_CNV_BAF_GENS.<tumor or normal>.baf.zoom.bed.gz
<meta.id>_<meta.sample_type>_CNV_COV_GENS.bed.gz

I suggest you modify to:
<meta.id>_<meta.sample_type>_CNV_BAF_GENS.bed.gz
<meta.id>_<meta.sample_type>_CNV_COV_GENS.bed.gz

So the extension should be only *.bed.gz, and the order of the elements correct. 

Modify this section and all other files that need to be changed

docs/output.md
```diff
@@ -73,10 +73,16 @@ This process produce an interactive CNV report.
...
+  - `<meta.id>_<meta.sample_type>_CNV_COV_GENS.bed.gz.tbi`: index file for the gzipped bed with coverage levels.
 
 </details>
 
 [`RMARKDOWNNOTEBOOK`](https://github.com/rstudio/rmarkdown) - used to generate a custom interactive CNV report using the RMARKDOWN script `assets/cnv_report.Rmd`.
+`PREPARE_AMBER_FOR_GENS` - Custom script that takes output from Oncoanalysers [`AMBER`](https://github.com/hartwigmedical/hmftools/tree/master/amber) VCF and converts the B-allele frequency levels into a GENS compatible zoom level.
+`PREPARE_COBALT_FOR_GENS` - Custom script that takes output from Oncoanalysers [`COBALT`](https://github.com/hartwigmedical/hmftools/tree/master/cobalt) VCF and converts the coverage levels into GENS compatible levels.
```
fevac, line 85: I'm a bit confused on why is this in the output documentation

docs/output.md
kristinebilgrav: ~~Currently they are tsv files. Should they be bed? This was what they were when Mathias handed it over.~~
Edit: will make into bed file!

docs/output.md
kristinebilgrav: I can update to bed file ending. 
Regarding the naming, for the BAF files, currently the sample name is `${meta.id}_CNV_BAF_GENS` and then the script adds whether its normal or tumor, which makes the naming like so;
`${meta.id}_CNV_BAF_GENS.normal/tumor.bed.gz`. 
This because the input file is a case file and the script was designed to do analysis of tumor and normal at the same time (script will skip normal if tumor_only). Is this an okay file name? Otherwise I think we can add an issue to refactor the script to a different logic perhaps? 🤔

docs/output.md
kristinebilgrav: I am not sure actually, but having the tools there was a part of the template, so I filled it out.

bin/prepare_amber_for_gens.py
kristinebilgrav: Yes, the plan is to update when we update the script. We want versions in the main.nf anyway, and the alternative was to write a value there (like here https://github.com/Clinical-Genomics/oncorefiner/pull/122#discussion_r3665034782) , but that would be more difficult to remember to update when updating the script. This way its easier to remember to update.

bin/prepare_amber_for_gens.py
kristinebilgrav: Added

bin/prepare_cobalt_for_gens.py
kristinebilgrav: I feel like @mathiasbio would be better at clarifying this since he wrote it and I am not sure. Could you help us here?

bin/prepare_cobalt_for_gens.py
kristinebilgrav: see above

bin/prepare_cobalt_for_gens.py
mathiasbio: Sorry yes! I can explain. So the different thing here compared to the GENS files we're working with from other pipelines, is that there we are merging multiple small regions (100bp in size) into progressively larger regions to implement the zoom functionality. Here from cobalt we already only have the segments, which when viewed in GENS would end up being 1 lonely little point. So to make it more viewable, we're basically expanding the segment by dividing it up into these chunks.

So if a segment is let's say 12 000 bp long, and we have a spacing of 5 000 we couldn't evenly divide the chunks up into 3 chunks of 5 000 bp in size. So instead when we divide:

n_points = max(1, math.ceil(length / spacing))

Like: 12000 / 5000 we get 2.4, and rounded up we get 3 chunks. Then we divide the segment into 3 chunks of 4000 bp in size instead. So it's basically that we're trying to achieve roughly segments of that size. 

The alternative is to take 3 segments again, but doing:

5000bp + 5000bp + 2000bp, but I think that's not as neat.

Basically this means that there will be slightly different spacing between the chunks for different segments, but I don't think it matters that much! At least as we get closer to the maximum zoom-level 1000bp, the resolution should be quite similar regardless.

bin/prepare_cobalt_for_gens.py
mathiasbio: I wouldn't say that we're generating log2ratios, we're just subdividing segments and reusing the log2ratios that cobalt has already calculated fro the segment

PR #120, opened 2026-06-02
feat: add `PREPARE_AMBER_FOR_GENS`local module

mathiasbio:
A part of https://github.com/Clinical-Genomics/MTP-oncoflow/issues/95

Main feature branch: https://github.com/Clinical-Genomics/oncorefiner/pull/119

### Added

- Local module `PREPARE_AMBER_FOR_GENS` local module along with tests 

The script adapts the amber BAF file to a GENS accepted BAF zoom file according to given levels of zoom


modules/local/prepare_amber_for_gens/main.nf
```diff
@@ -0,0 +1,49 @@
...
+    input:
+    tuple val(meta), path(amber_baf_tsv), val(sampletype)
+
+    output:
+    tuple val(meta), path("*tumor.baf.zoom.tsv.gz"), emit: tumor_tsv
+    tuple val(meta), path("*normal.baf.zoom.tsv.gz"), emit: normal_tsv, optional: true
+    tuple val(meta), path("*tumor.baf.zoom.tsv.gz.tbi"), emit: tumor_tbi
+    tuple val(meta), path("*normal.baf.zoom.tsv.gz.tbi"), emit: normal_tbi, optional: true
+    tuple val("${task.process}"), val('prepare_amber_for_gens'), val('1.0'), topic: versions, emit: versions_amber_for_gens
+    // WARN: Version information not provided by tool on CLI. Please update version string above when bumping container versions.
```
ramprasadn: Do you think we should record version strings of other tools in the container?

kristinebilgrav: Hm, good point. I can give it a try!

kristinebilgrav: Updated! Let me know what you think.

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,119 @@
...
+@click.option(
+    "--input-file",
+    required=True,
+    type=click.Path(exists=True),
+    help="Input BAF TSV file",
+)
+@click.option(
+    "--sample-type",
+    show_default=True,
+    help="Sample type (e.g., tumor_only or tumor_normal)",
```
mathiasbio:
```suggestion
    type=click.Choice(["tumor_only", "tumor_normal"]),
    help="Analysis type",
```

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,119 @@
...
+
+@click.command()
+@click.option(
+    "--input-file",
+    required=True,
+    type=click.Path(exists=True),
+    help="Input BAF TSV file",
+)
+@click.option(
+    "--sample-type",
```
mathiasbio:
```suggestion
    "--analysis-type",
```

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,119 @@
...
+    show_default=True,
+    help="Output file prefix for BAF",
+)
+
+@click.version_option(__version__, prog_name="prepare_amber_for_gens")
+
+def main(
+    input_file: click.Path,
+    output_file_prefix: click.STRING,
+    sample_type: click.STRING,
```
mathiasbio:
```suggestion
    analysis_type: click.STRING,
```

bin/prepare_amber_for_gens.py
```diff
@@ -0,0 +1,119 @@
...
+
+    write_baf_file(
+        df=df,
+        baf_column="tumorBAF",
+        output_path=f"{output_file_prefix}.tumor.baf.zoom.tsv",
+        levels=DEFAULT_LEVELS,
+    )
+    click.echo(f"Wrote tumor output: {output_file_prefix}.tumor.baf.zoom.tsv")
+
+    if "normal" in sample_type:
```
mathiasbio:
```suggestion
    if "normal" in analysis_type:
```

bin/prepare_amber_for_gens.py
mathiasbio: I realise I'm the one originally that named it sample-type 🥲 but I don't know. It seems more consistent with the naming in CG overall to refer to this as an analysis or case rather than sample. But I'm ok keeping it like this too, and we can potentially refactor it in a later PR!

bin/prepare_amber_for_gens.py
mathiasbio:
```suggestion
    type=click.Choice(["tumor_only", "tumor_normal"]),
    help="Sample type (tumor_only or tumor_normal)",
```
Feel free to reject the other comments renaming sample_type to analysis_type. But here's a suggestion just to make it a bit more robust to typos. So that we don't accidentally write norml or something 😂

modules/local/prepare_amber_for_gens/tests/main.nf.test
```diff
@@ -0,0 +1,133 @@
...
+                    [id: 'test_amber_gens', case_id: 'subject_a'],
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.amber.baf.tsv.gz', checkIfExists: true),
+                    ])
+                input[1] = 'tumor_normal'
+                """
+            }
+        }
+
+        then {
+            def output_tsv = process.out.findAll { key, value -> key.endsWith('_tsv')}
```
mathiasbio, line 28: Would it make sense to assert that the normal file is produced? And inversely that it is not in the tumor only case?

modules/local/prepare_amber_for_gens/main.nf
```diff
@@ -0,0 +1,53 @@
...
+        fi
+    done
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    sampletype=${sampletype}
+    if [[ $sampletype == *"normal"* ]]; then
+        echo | gzip > ${prefix}.${sampletype}.normal.baf.zoom.tsv.gz
```
mathiasbio: Do we need the sampletype in the file name? It seems a bit repetitive.

modules/local/prepare_amber_for_gens/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+  - sampletype:
+      type: string
+      description: Sample type, e.g. "tumor_only" or "tumor_normal"
+output:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing case information
+        e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor,normal', sex: 'female']`
+  - tsv_tumor:
```
mathiasbio:
```suggestion
  - tumor_tsv:
```

modules/local/prepare_amber_for_gens/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+      type: file
+      description: bgzipped and tabix-indexed AMBER BAF TSV zoom file for tumor samples
+      pattern: "*.baf.zoom.tsv.gz"
+  - tsv_normal:
```
mathiasbio:
```suggestion
  - normal_tsv:
```

modules/local/prepare_amber_for_gens/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+      type: file
+      description: bgzipped and tabix-indexed AMBER BAF TSV zoom file for normal samples
+      pattern: "*.baf.zoom.tsv.gz"
+      optional: true
+  - tbi_tumor:
```
mathiasbio:
```suggestion
  - tumor_tbi:
```

modules/local/prepare_amber_for_gens/meta.yml
```diff
@@ -0,0 +1,62 @@
...
+      type: file
+      description: tabix index for AMBER BAF TSV zoom file for tumor samples
+      pattern: "*.baf.zoom.tsv.gz.tbi"
+  - tbi_normal:
```
mathiasbio:
```suggestion
  - normal_tbi:
```

modules/local/prepare_amber_for_gens/meta.yml
```diff
@@ -0,0 +1,62 @@
+name: prepare_amber_for_gens
+description: Convert column mean in file *tumor.amber.baf.tsv.gz into GENS zoom-level data.
+keywords:
+  - baf
+  - amber
+  - tabix
+  - bgzip
+tools:
+  - "python":
```
mathiasbio:
```suggestion
  - "prepare_amber_for_gens.py":
```

modules/local/prepare_amber_for_gens/main.nf
```diff
@@ -0,0 +1,53 @@
+process PREPARE_AMBER_FOR_GENS {
+    tag "$meta.id"
+    tag "process_single"
```
mathiasbio:
```suggestion
```
From what I heard only one tag is actually used

mathiasbio: Sorry, see comment here: https://github.com/Clinical-Genomics/oncorefiner/pull/122/changes#r3728717333

bin/prepare_amber_for_gens.py
kristinebilgrav: I think it can have been me as well. Regardless, I agree that its too similar to our meta sample_type, and this is just a custom value to differentiate the files, so better to keep different!

bin/prepare_amber_for_gens.py
kristinebilgrav: Nice, thanks! :D

modules/local/prepare_amber_for_gens/tests/main.nf.test
kristinebilgrav: The normal file is asserted. If analysis_type is tumor_normal, both tumor and normal files are produced. And for tumor_only, the tumor file is the only one that is produced, so they are both in the snapshot. There is no specific assertion for checking if normal is *not* produced there, but I dont exactly know how to do that 😅

modules/local/prepare_amber_for_gens/main.nf
kristinebilgrav: I agree, have removed that!

modules/local/prepare_amber_for_gens/tests/main.nf.test
mathiasbio: I see! So it's just a matter of seeing if the snapshot is different, suddenly a file disappears from the snapshot and then we'd investigate? I'm good with that 🙏

kristinebilgrav: I think so, but there could be a more sophisticated way I dont know about 😅

modules/local/prepare_amber_for_gens/main.nf
```diff
@@ -0,0 +1,52 @@
...
+            bgzip -f "\$f"
+            tabix -f -s 1 -b 2 -e 3 "\$f.gz"
+        fi
+    done
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    if [[ $analysis_type == *"normal"* ]]; then
```
mathiasbio:
```suggestion
    if [[ $analysis_type == *normal* ]]; then
```
Tiny thing, feel free to ignore!

PR #122, opened 2026-06-03
feat: add local module `prepare_cobalt_for_gens`

mathiasbio:
A part of: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/95

Merge to feature branch: 
https://github.com/Clinical-Genomics/oncorefiner/pull/119

Short description of the script:

```
DEFAULT_SPACING = {
    "o": 50_000,
    "a": 20_000,
    "b": 10_000,
    "c": 5_000,
    "d": 1_000,
}
```

These correspond to the different zoom levels of GENS. Where o is the most zoomed out and d is the most zoomed in. Usually in GENS the input file is a file that contains 100bp window bins with log2 coverage ratios, and the zoom level decides how many to merge together and average. But in this case we have the opposite problem, as the cobolt pcf file contains segments that are typically quite large! So if we were to not subdivide the regions we would have very sparse data, which might work in the most zoomed out level (with 1 datapoint per segment). But if we zoomed in it would be hard to see the segment at all.

...

### Added

- Local module `prepare_cobalt_for_gens`, which takes the cobalt file and adapts it like above to comply with GENS
- Added test for local module `prepare_cobalt_for_gens`,


bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python3
+
+import math
```
kristinebilgrav, line 3: Built-in module, did not include this in the container

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,38 @@
+process PREPARE_COBALT_FOR_GENS {
+    tag "$meta.id"
```
imsarath, line 2: Missing resource label - "process_single" ?

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    tuple val(meta), path(cobalt_pcf)
+
+    output:
+    tuple val(meta), path("*${meta.sample_type}.bed.gz"),  emit: bed
+    tuple val(meta), path("*${meta.sample_type}.bed.gz.tbi"), emit: tbi
+    tuple val("${task.process}"), val('prepare_cobalt_for_gens'), val('1.0'), topic: versions, emit: versions_cobalt_for_gens
+    // WARN: Version information not provided by tool on CLI. Please update version string above when bumping container versions.
+
+    script:
+    def prefix = meta.id ?: "sample"
```
imsarath: Consider the following nf-core convention, so output naming is consistent and overridable via modules config file.
```suggestion
    def args        = task.ext.args ?: ''
    def prefix      = task.ext.prefix ?: "${meta.id}"
```

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+    'https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/68/68a2ae938ce2d285498e0a81233acf6bd423ac910c3fe6fc68f44322559b1187/data' :
+    'community.wave.seqera.io/library/tabix_click_pandas_py-bgzip_python:4d1c106d561e2b25' }"
+
+    input:
+    tuple val(meta), path(cobalt_pcf)
+
+    output:
+    tuple val(meta), path("*${meta.sample_type}.bed.gz"),  emit: bed
```
imsarath:
```suggestion
    tuple val(meta), path("*.bed.gz"),  emit: bed
    tuple val(meta), path("*.bed.gz.tbi"), emit: tbi
```

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    tuple val(meta), path("*${meta.sample_type}.bed.gz.tbi"), emit: tbi
+    tuple val("${task.process}"), val('prepare_cobalt_for_gens'), val('1.0'), topic: versions, emit: versions_cobalt_for_gens
```
imsarath: Version is hardcoded to val('1.0'); Would it be possible to add versioning in the Python script and make it dynamic eg: `eval("script.py --version | .. ")`?

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        --input-file ${cobalt_pcf} \\
+        --output-file ${prefix}.${meta.sample_type}.bed
+
+    bgzip ${prefix}.${meta.sample_type}.bed
+
+    tabix ${prefix}.${meta.sample_type}.bed.gz
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
```
imsarath, line 35: Different prefix logic for script and stub; it's better to maintain uniformity.

modules/local/prepare_cobalt_for_gens/meta.yml
```diff
@@ -0,0 +1,49 @@
...
+  - "tabix":
+      description: "Index tab-delimited genome position files."
+      homepage: "http://www.htslib.org/doc/tabix.html"
+      identifier: tabix
+input:
+  - meta:
+      type: map
+      description: |
+        Groovy Map containing case information
+        e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor,normal', sex: 'female']`
```
imsarath:
```suggestion
        e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor (or) normal', sex: 'female']`
```

modules/local/prepare_cobalt_for_gens/tests/main.nf.test
```diff
@@ -0,0 +1,210 @@
...
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    process.out,
+                ).match() }
```
imsarath, line 85: Sanitize process output to create clean snapshots

```suggestion
               { assert snapshot(sanitizeOutput(process.out)).match() }
```

modules/local/prepare_cobalt_for_gens/tests/main.nf.test.snap
```diff
@@ -0,0 +1,410 @@
...
+                ]
+            }
+        ],
+        "timestamp": "2026-07-28T09:37:15.746937",
+        "meta": {
+            "nf-test": "0.9.5",
+            "nextflow": "26.04.0"
+        }
+    },
+    "Prepare cobalt files for GENS, normal (tumor-only data)": {
```
imsarath: Missing output from the normal sample?

modules/local/prepare_cobalt_for_gens/tests/main.nf.test
kristinebilgrav: This is great, thank you!

modules/local/prepare_cobalt_for_gens/tests/main.nf.test.snap
kristinebilgrav: Yes, for the tumor-only, the test is to verify that it will work to provide an empty channel since there will be no normal file. Not sure if I should provide channel.empty() or null. But probably good to test that it does not crash?

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Perhaps, what does this label indicate?

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: I think I need the sample_type here as I specify that in line 24 below?

modules/local/prepare_cobalt_for_gens/main.nf
imsarath: It is just to allocate how many resources this module requires - [conf/base.config](https://github.com/Clinical-Genomics/oncorefiner/blob/dev/conf/base.config)

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Oh, good suggestion. Although I think I would need some help in implementing it in the naming of the output file below.

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: good catch, thanks!

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Ah, thanks !!

modules/local/prepare_cobalt_for_gens/main.nf
imsarath: You don't need ${meta.sample_type} in the glob  *.bed.gz, it already matches the process output. The `sample_type` in the glob would only be needed if the process emitted multiple .bed.gz files and you had to distinguish between them. My recommendation is to avoid referencing meta.sample_type inside the module so it stays as general as possible; if you need control over the output naming, the cleaner approach is to set the prefix via task.ext.prefix in conf/modules.config rather than hardcoding meta fields in the module.

modules/local/prepare_cobalt_for_gens/tests/main.nf.test.snap
imsarath: Now looking back at this, If the input is an empty channel, the process simply isn't invoked, so there's nothing for the module to crash on. Ideally, if I am not wrong, the tumor-normal or tumor-only case is filtered upstream in the workflow, so we don't really need to cover it here. I'd recommend keeping the test focused on this module alone.

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: I see, thanks! I removed this from the output statement. 
Regarding the use of sample_type, I see your point! I initially did it to ensure the files were not overwritten, since the process is performed for both tumor and normal in paralell, and there is a risk that they will be named the same. I still see this risk though. I added the  ${meta.sample_type}  in the prefix definition. What do you think of this approach?

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Hm, good point. Will look into it and see what I can do.

modules/local/prepare_cobalt_for_gens/main.nf
imsarath: Good point! The collision risk is real if tumor and normal share meta.id, so sample_type needs to be in the filename. My only suggestion is to drive it from config rather than hardcoding it in the module. Keep the module generic: `def prefix = task.ext.prefix ?: "${meta.id}" `

with globs `*.bed.gz / *.bed.gz.tbi`, and set the naming in conf/modules.config:
```
withName: PREPARE_COBALT_FOR_GENS {
     ext.prefix = { "${meta.id}_${meta.sample_type}" }
}

```
Same collision-safe names, but sample_type stays out of the module so it's reusable.

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,40 @@
...
+    script:
+    def args   = task.ext.args ?: ''
+    def prefix = task.ext.prefix ?: "${meta.id}.${meta.sample_type}"
+
+    """
+    prepare_cobalt_for_gens.py \\
+        --input-file ${cobalt_pcf} \\
+        --output-file ${prefix}.bed
+
+    bgzip ${prefix}.bed
```
imsarath, line 29: Minor nf-core convention note: chaining multiple tools (bgzip + tabix) in a single module is generally discouraged - each module should ideally wrap one tool. If feasible, consider having the Python script emit a plain .bed and handle compression/indexing with the dedicated `nf-core/htslib/bgziptabix` module downstream.

That said, I'm not fully certain here; it'd be good to get confirmation from a more experienced nf dev before acting on this.

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Updated!

modules/local/prepare_cobalt_for_gens/tests/main.nf.test.snap
kristinebilgrav: This is true, the test will be on a subworkflow/workflow level, not on module, so will remove and only do for tumor and normal.

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: I still see the risk in it being forgotten to be included in the config somewhere down the line, but I guess its better not to hardcode it, so the metas can be different, so applied your suggestion. What do you think @mathiasbio, do you have another idea or input?

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: What do you think @ramprasadn ?

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: I realize that with the prefix being configurable with the config, and the script not really accepting other parameters, I dont see the use case for the args?

imsarath: Yes, you're right! We don't need the `args` option right now since the Python script only uses input and output. We can always add `args` option later if needed.

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,42 @@
+process PREPARE_COBALT_FOR_GENS {
+    tag "$meta.id"
+    tag "process_single"
```
mathiasbio:
```suggestion
    label "process_single"
```
Maybe it should be this instead? And same in the other PR

modules/local/prepare_cobalt_for_gens/main.nf
mathiasbio: Hmm. Here's again I think where the sample_type name for the variable may be confusing. 

To me when I hear sample_type (and again my fault originally for calling it this 😂 ) I hear "tumor" or "normal", I don't hear tumor_only, or tumor_normal, which I think is what this variable is: https://github.com/Clinical-Genomics/oncorefiner/pull/112/changes#diff-7ad93860b53d91cc603737c90612d6533ce53edf882eda504ddd476df97c9a2f

I think the name tumor or normal would be a lot more fitting. But that's probably a different question.

I don't know how generally used this module will be in the end, but I still like the idea of not hardcoding these things in the module. But I don't know if I'm good enough at nextflow to understand the risks.

modules/local/prepare_cobalt_for_gens/main.nf
```diff
@@ -0,0 +1,42 @@
...
+    output:
+    tuple val(meta), path("*.bed.gz"),  emit: bed
+    tuple val(meta), path("*.bed.gz.tbi"), emit: tbi
+    tuple val("${task.process}"), val('python'), eval("python --version | sed '1!d; s/^.*python //'"), topic: versions, emit: versions_python
+    tuple val("${task.process}"), val('bgzip'), eval("bgzip --version | sed '1!d; s/^.*bgzip //'"), topic: versions, emit: versions_bgzip
+    tuple val("${task.process}"), val('tabix'), eval("tabix --version | sed '1!d; s/^.*tabix //'"), topic: versions, emit: versions_tabix
+    tuple val("${task.process}"), val('prepare_cobalt_for_gens'), eval("prepare_cobalt_for_gens.py --version | sed '1!d; s/^.*prepare_cobalt_for_gens //'"), topic: versions, emit: versions_cobalt_for_gens
+
+    script:
+    def args   = task.ext.args ?: ''
```
mathiasbio: If we don't use this, maybe it should be removed

modules/local/prepare_cobalt_for_gens/meta.yml
```diff
@@ -0,0 +1,49 @@
+name: prepare_cobalt_for_gens
+description: Convert column mean in file .cobalt.ratio.pcf into GENS zoom-level data.
+keywords:
+  - cov
+  - cobalt
+  - tabix
+  - bgzip
+tools:
+  - "python":
```
mathiasbio:
```suggestion
  - "prepare_cobalt_for_gens.py":
```

bin/prepare_cobalt_for_gens.py
```diff
@@ -0,0 +1,125 @@
...
+    out_df = pd.DataFrame(rows)
+    out_df.to_csv(output_path, sep="\t", header=False, index=False)
+
+
+@click.command()
+@click.option(
+    "--input-file",
+    required=True,
+    type=click.Path(exists=True),
+    help="Input segment PCF file from Cobolt. Can be plain text or .gz.",
```
mathiasbio:
```suggestion
    help="Input segment PCF file from Cobalt. Can be plain text or .gz.",
```

modules/local/prepare_cobalt_for_gens/main.nf
mathiasbio: I don't know much about this, but I think it's correct. However, I don't know personally how much this module is intended to be shared. It's a very custom solution that I'm not sure we're even going to put into production. But Eva may not agree with me there 😂

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Ah, good catch!

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: True, I will change the variable to analysis_type or similar in the other PR since its too similar. But here I think its the meta, so here its actually tumor or normal. (which makes the other sampletype variable even more confusing haha). 
But regardless, I will keep it out of the module to avoid hard-coding then!

modules/local/prepare_cobalt_for_gens/main.nf
kristinebilgrav: Hehe, okay, I think I will keep it for now then, and we can lift those parts out later 😅

PR #123, opened 2026-06-08
test: update subworkflow tests for `PREPARE_REFERENCES` to check workflow output

kristinebilgrav:
### Fixed

- Updated so the test for untarring of vep cache checks the `workflow.out`
- Updated snapshot


subworkflows/local/prepare_references/tests/main.nf.test
beatrizsavinhas: No need to keep this structure for this test. If the input doesn't need untarring, the subworkflow merely retrieves the same path that is given as input.
I suggest removing the check for the output directories and checking instead that `UNTAR_VEP_CACHE` is not run!

beatrizsavinhas: Same as above!

beatrizsavinhas: We had talked about removing this as well, right?

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -86,7 +86,7 @@ nextflow_workflow {
 
         then {
             // Check untarred directory and files
-            def output_directories_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)
+            def output_directories_files = getAllFilesFromDir(workflow.out.vep_resources[0][0], relative: true, includeDir: true)
```
beatrizsavinhas: Can you create a variable to avoid code duplication with `workflow.out.vep_resources[0][0]` like we discussed yesterday?

beatrizsavinhas (changes requested):

beatrizsavinhas:
Also, the prefix for the PR title should be "test"!

subworkflows/local/prepare_references/tests/main.nf.test
kristinebilgrav: True! When you mention checking that `UNTAR_VEP_CACHE` is not run, how do we check that?

fellen31: You could have a look at https://github.com/askimed/nf-test/pull/355!

beatrizsavinhas, line 10:
```suggestion

    topics "versions"
```

subworkflows/local/prepare_references/tests/main.nf.test
```diff
@@ -23,16 +23,12 @@ nextflow_workflow {
         }
 
         then {
-            // Check untarred directory and files
-            def output_directories_files = getAllFilesFromDir(params.outdir, relative: true, includeDir: true)
 
             assertAll(
                 { assert workflow.success },
                 { assert snapshot(
                     // Check the name of the output directory
```
beatrizsavinhas, line 33:
```suggestion
                    // Check topic versions to see what processes were run
                    topics,
                    // Check the name of the output directory
```

beatrizsavinhas: This can be applied for all tests!

subworkflows/local/prepare_references/tests/main.nf.test
beatrizsavinhas: See my suggestions below for implementing this.

subworkflows/local/prepare_references/tests/main.nf.test
kristinebilgrav: aah, I see! Thank you! 🙏

beatrizsavinhas (approved):
Nice to get this cleaned up! 🧹

PR #126, opened 2026-06-09
feat: CNV report using nf-core rmarkdownotebook module

fevac:
Part of: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/34

### Added

- nf-core rmarkdownotebook module
- PROCESS_CNV subworkflow with CNV report generation step
    - subworkflow test and snapshot for process_cnvs

### Changed

- Updated tests with files for the subworkflow
- Updated snapshots for pipeline tests


subworkflows/local/process_cnvs/main.nf
```diff
@@ -0,0 +1,62 @@
...
+
+workflow PROCESS_CNVS {
+
+    take:
+    ch_cnv_gene_tsv                 // channel: [optional]  [val(meta), path(tsv)]
+    ch_cnv_segment_tsv              // channel: [optional]  [val(meta), path(tsv)]  
+    
+    main:
+    // Path to report template (Rmd file)
+    def cnv_report_template = file("${projectDir}/assets/cnv_report.Rmd", checkIfExists: true)
```
fevac, line 28: this could be a parameter that by default uses this template but alternatively a different template can be given, which could allow for easier testing and customization in dev. But out of scope

subworkflows/local/process_cnvs/main.nf
```diff
@@ -0,0 +1,62 @@
...
+
+    // Join the two distinct file channels together based on the meta.id
+    ch_combined_inputs = ch_cnv_gene_tsv.join(ch_cnv_segment_tsv)
+
+    // Prepare the exact tuple format expected by the module's input:
+    // Input 1: tuple val(meta), path(notebook)
```
fevac: called notebook in the module, and here refers to the cnv report rmarkdown template

subworkflows/local/process_cnvs/main.nf
```diff
@@ -0,0 +1,62 @@
...
+    // -------------------------------------------------------------
+    RMARKDOWNNOTEBOOK (
+        ch_report_inputs.map { it[0] }, // tuple val(meta), path(notebook)
+        ch_report_inputs.map { it[1] }, // val parameters
+        ch_report_inputs.map { it[2] }  // path input_files
+    )
+
+    emit:
+    html_report = RMARKDOWNNOTEBOOK.out.report       // channel: [ val(meta), path(*.html) ]
+    versions    = RMARKDOWNNOTEBOOK.out.versions     // channel: [ path(versions.yml) ]
```
fevac: need to check versions and modify accordingly

assets/cnv_report.Rmd
kristinebilgrav, line 1: The space changes are precommit automatic changes

assets/cnv_report.Rmd
```diff
@@ -1,6 +1,6 @@
 ---
 title: "CNV Report"
-date: "`r format(Sys.time(), '%Y-%m-%d %H:%M %Z')`"
+date: "`r format(Sys.time(), \"%Y-%m-%d %H:%M %Z\")`"
```
kristinebilgrav, line 3: Was failing without this change

subworkflows/local/process_cnvs/tests/main.nf.test.snap
```diff
@@ -0,0 +1,46 @@
...
+                [
+                    {
+                        "id": "CNV_REPORT"
+                    },
+                    "CNV_REPORT_cnv_report.html"
+                ]
+            ],
+            {
+                "versions": [
+
```
kristinebilgrav: The nf-core module still has the old way of adding versions, so the versions part will be empty until this is fixed on their side

main.nf
```diff
@@ -142,6 +144,12 @@ workflow CLINICALGENOMICS_ONCOREFINER {
     ch_sv_dbs            = val_svdb_query_dbs    ? channel.fromPath(val_svdb_query_dbs)
                                                  : channel.empty()
 
+    // Input for CNV report
+    ch_cnv_gene_tsv      = val_cnv_gene_tsv      ? channel.fromPath(val_cnv_gene_tsv).map { it -> [[id:'CNV_REPORT'], it] }.collect()
+                                                 : channel.empty()
+    ch_cnv_segment_tsv   = val_cnv_segment_tsv   ? channel.fromPath(val_cnv_segment_tsv).map { it -> [[id:'CNV_REPORT'], it] }.collect()
+                                                 : channel.empty()
```
kristinebilgrav, line 150: Should this be written in the new way with the metas?

modules/nf-core/rmarkdownnotebook/main.nf
```diff
@@ -0,0 +1,129 @@
...
+    tag "$meta.id"
+    label 'process_low'
+
+    //NB: You likely want to override this with a container containing all required
+    //dependencies for your analysis. The container at least needs to contain the
+    //yaml and rmarkdown R packages.
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine in ['singularity', 'apptainer'] && !task.ext.singularity_pull_docker_container ?
+        'oras://community.wave.seqera.io/library/r-base_r-dt_r-readr_r-rmarkdown:8612e5aa5384f180' :
+        'community.wave.seqera.io/library/r-base_r-dt_r-readr_r-rmarkdown:5c3b8868d1024810' }"
```
fevac, line 11: Note: This was overwritten to support modules needed by the report template

main.nf
fevac: I think so. They are the the case level

modules/nf-core/rmarkdownnotebook/main.nf
fellen31: Please use the nf-core patch functionally: https://nf-co.re/docs/nf-core-tools/cli/modules/patch.

subworkflows/local/process_cnvs/tests/main.nf.test.snap
beatrizsavinhas: In that case, the subworkflow needs to include logic to mix the module version in the version topic channel (the old way).

fellen31: More or less equal amount of effort as updating the nf-core module IMO.

beatrizsavinhas: Agreed! I would vote for that too!

kristinebilgrav: I agree, and I do plan on opening an issue and fixing it, but IMO I think we can still move on with this PR and then updating it later once its fixed?

beatrizsavinhas: That I disagree with because we shouldn't be adding a module without recording the software versions.

fevac: Do you require a template update to be able to add an updated nf-core module? or can you add it directly as long as the changes are merged in master? 

I personally think it's ok to add a nf-core module as-is and temporarily skipping the version, as long as the issue to tackle this gets created and documented and is worked in parallel in a specific PR.  To me this is a `pre-existing issue` that can be separated from the main issue.

fevac: To me both are valid options but I'd strongly favour separating this if a nf-core release is required to put the fix into place

beatrizsavinhas: No, the template is independent from the modules. Modules can be updated super easily with `nf-core modules update`.

I would not say this is an issue, per se. It's just a limitations of the module currently so we have to work with it and the fixes are very simple.
Mixing the module version with the topic channel versions is just one line of code. Fixing the nf-core module is changing one line of code and a PR.

modules/nf-core/rmarkdownnotebook/main.nf
kristinebilgrav: ✅

subworkflows/local/process_cnvs/tests/main.nf.test.snap
kristinebilgrav: Updated the module here: https://github.com/nf-core/modules/pull/12193

subworkflows/local/process_cnvs/main.nf
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+//
+// MODULE: Installed directly from nf-core/modules
+//
+include { RMARKDOWNNOTEBOOK } from '../../../modules/nf-core/rmarkdownnotebook/main'
+
+/*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    RUN PROCESS_SNVS WORKFLOW
```
kristinebilgrav, line 14:
```suggestion
    RUN PROCESS_CNVS WORKFLOW
```

subworkflows/local/process_cnvs/main.nf
```diff
@@ -0,0 +1,54 @@
...
+            def data_files = [cnv_gene_tsv, cnv_segment_tsv]
+
+            return [ [meta, cnv_report_template], r_params, data_files ]
+        }
+
+    RMARKDOWNNOTEBOOK (
+        ch_report_inputs.map { it[0] }, // tuple val(meta), path(notebook)
+        ch_report_inputs.map { it[1] }, // val parameters
+        ch_report_inputs.map { it[2] }  // path input_files
+    )
```
ramprasadn: Would this work?
```suggestion
    .multiMap { notebook_tuple, r_params, data_files ->
        notebook   : notebook_tuple
        parameters : r_params
        input_files: data_files
    }

RMARKDOWNNOTEBOOK(
    ch_report_inputs.notebook,
    ch_report_inputs.parameters,
    ch_report_inputs.input_files,
)```

Comment (ramprasadn) [reply subworkflows/local/process_cnvs/main.nf]: Or this?

```
ch_report_inputs = ch_cnv_gene_tsv
    .join(ch_cnv_segment_tsv)
    .multiMap { meta, cnv_gene_tsv, cnv_segment_tsv ->
        notebook   : [meta, cnv_report_template]
        parameters : [cnv_gene: cnv_gene_tsv.name, cnv_segment: cnv_segment_tsv.name]
        input_files: [cnv_gene_tsv, cnv_segment_tsv]
    }
RMARKDOWNNOTEBOOK(
    ch_report_inputs.notebook,
    ch_report_inputs.parameters,
    ch_report_inputs.input_files)
```

Comment (ramprasadn) [reply subworkflows/local/process_cnvs/main.nf]: Implicit `it` is deprecated in Nextflow's strict syntax. So even if my suggestions don't work, it would be good to declare the parameters explicity inside closures.

File: CHANGELOG.md [hunk 10]
```diff
@@ -79,6 +81,8 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#145](https://github.com/Clinical-Genomics/oncorefiner/pull/145) Small fixes to `GENERATE_CYTOSURE_FILES` tests from #135: updated config regex to include the fully qualified name of the process in order for the settings to be applied with the highest priority and removed parameter dependency from each test by moving the values to the config file instead.
 - [#149](https://github.com/Clinical-Genomics/oncorefiner/pull/149) Updated test data base path to remove unused column in SVDB files
 - [#151](https://github.com/Clinical-Genomics/oncorefiner/pull/151) Updated test data base path to include CNV files
+- [#126](https://github.com/Clinical-Genomics/oncorefiner/pull/126) Updated config test files to include parameters and files for `PROCESS_CNVS`
+- [#126](https://github.com/Clinical-Genomics/oncorefiner/pull/126) Updated pipeline test snapshot to accomodate for `PROCESS_CNVS`
```
beatrizsavinhas: Don't think this is needed. It's implied when a new module/subworkflow that produces new output files is added.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#135](https://github.com/Clinical-Genomics/oncorefiner/pull/135) Added test config for `GENERATE_CYTOSURE_FILES`.
 - [#124](https://github.com/Clinical-Genomics/oncorefiner/pull/124) Added metadata parameters `case_id`, `sample_id_tumor` and `sample_id_normal`.
 - [#124](https://github.com/Clinical-Genomics/oncorefiner/pull/124) Added `channelFromMetaAndPath` custom channel factory.
+- [#126](https://github.com/Clinical-Genomics/oncorefiner/pull/126) Added the nf-core module `rmarkdownnotebook`
```
```suggestion
- [#126](https://github.com/Clinical-Genomics/oncorefiner/pull/126) Added the nf-core module `rmarkdownnotebook`.
```
Check all lines added in this PR.

subworkflows/local/process_cnvs/main.nf
beatrizsavinhas: Highly recommend this!

CHANGELOG.md
beatrizsavinhas, line 1: There's no mention of what is actually being run with the `RMARKDOWNNOTEBOOK` module.

CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
...
+- [#126](https://github.com/Clinical-Genomics/oncorefiner/pull/126) Added the subworkflow `PROCESS_CNVS` along with subworkflow test.
```
beatrizsavinhas, line 44: This is the main feature of this PR. Please update the title of the PR accordingly.

assets/cnv_report.Rmd
beatrizsavinhas: Suggest applying fixes like this in another PR (or even in the feature branch itself).

tests/.nftignore
```diff
@@ -11,3 +11,4 @@ multiqc/multiqc_data/multiqc_citations.txt
 pipeline_info/*.{html,json,txt,yml}
 vep/*.{html}
 alignments/*.{cram,crai}
+cnv/*.{html}
```
beatrizsavinhas, line 14: Change not mentioned in description and there's no explanation as to why this is needed (presumably because the report includes the date?). Perhaps good to add to chagelog as well?

subworkflows/local/process_cnvs/main.nf
beatrizsavinhas, line 1: It's a bit confusing why this is a subworkflow when only 1 module is called. Is the idea to add more functionality to here in the future?

subworkflows/local/process_cnvs/main.nf
kristinebilgrav: Thank you, I didnt know this would be deprecated. Both work, but I think your second suggestion is clean and nice, thank you!

subworkflows/local/process_cnvs/main.nf
kristinebilgrav: Yes, the plan is that in the future there will be more analysis here.

CHANGELOG.md
kristinebilgrav: That is because the script was added in a seperate PR, and will thus not be added here. Better to add more details in the feature branch.

tests/.nftignore
kristinebilgrav: Adding to description.

PR #129, opened 2026-06-15
fix: update extra args to quotes

kristinebilgrav:
### Fixed

- Updated extra arguments `extra_args_snv_vep` and `extra_args_sv_vep` to be '' instead of null as they were causing a failure
- `test_base.config` does not overwrite the default unless necessary


fevac (approved):
🚀

fevac (approved):
Please update the change log

PR #130, opened 2026-06-15
fix: output SNV `research_filtered_vcf` after annotation

beatrizsavinhas:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/84.

### Changed

- Output logic for `PROCESS_SNVS` to output the `research_filtered_vcf` file after annotation with vep. The file name for this output is now given by vep, i.e. with prefix `${meta.id}_vep`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#116](https://github.com/Clinical-Genomics/oncorefiner/pull/116) Updated `GENERATE_CYTOSURE_FILES` subworkflow test snapshot in sequence of [#110](https://github.com/Clinical-Genomics/oncorefiner/pull/110).
 - [#123](https://github.com/Clinical-Genomics/oncorefiner/pull/123) Changed the subworkflow test when untarring is needed in `PREPARE_REFERENCES` to check `workflow.out.vep_resources` instead of `params.outdir`
 - [#129](https://github.com/Clinical-Genomics/oncorefiner/pull/129) Changed the extra arguments `extra_args_*` to default to '' instead of null and removed empty `extra_args_snv_vep` in `test_base.config`
+- [#130](https://github.com/Clinical-Genomics/oncorefiner/pull/130) Changed output logic for `PROCESS_SNVS` to output the `research_filtered_vcf` file after annotation with vep.
 
 ### `Dependencies`
 
--- a/tests/tumor_normal.nf.test.snap
+++ b/tests/tumor_normal.nf.test.snap
@@ -61,8 +61,6 @@
                 "snv",
                 "snv/subject_a_clinical.vcf.gz",
                 "snv/subject_a_clinical.vcf.gz.tbi",
-                "snv/subject_a_research.vcf.gz",
-                "snv/subject_a_research.vcf.gz.tbi",
                 "snv/subject_a_vcfanno.vcf.gz",
                 "snv/subject_a_vcfanno.vcf.gz.tbi",
                 "snv/subject_a_vep.vcf.gz",
@@ -77,11 +75,11 @@
                 "vep/subject_a_svdbquery_vep_summary.html"
             ]
         ],
+        "timestamp": "2026-06-15T14:21:20.993278",
         "meta": {
-            "nf-test": "0.9.3",
-            "nextflow": "25.10.4"
-        },
-        "timestamp": "2026-06-08T09:56:55.147283"
+            "nf-test": "0.9.5",
+            "nextflow": "26.04.2"
+        }
     },
     "Tumor Normal analysis with '-profile test'": {
         "content": [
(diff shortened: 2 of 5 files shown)


tests/tumor_normal.nf.test.snap
beatrizsavinhas, line 67: The output `research_filtered_vcf` file will come from the last step of this subworkflow (before clinical filtering). That means that now it has the name set by VEP.
When adding optional steps like scoring in https://github.com/Clinical-Genomics/oncorefiner/pull/103, the output file name will be variable.
Is that something we are OK with?

fellen31: This is something you could potentially fix within the workflow publish section.

tests/tumor_normal.nf.test.snap
```diff
@@ -61,8 +61,6 @@
                 "snv",
                 "snv/subject_a_clinical.vcf.gz",
                 "snv/subject_a_clinical.vcf.gz.tbi",
-                "snv/subject_a_research.vcf.gz",
-                "snv/subject_a_research.vcf.gz.tbi",
```
fellen31, line 65: So this file is "removed" since it's now a duplicate of  "snv/subject_a_vep.vcf.gz"?

fellen31 (commented):
If you don't want to deal with https://github.com/Clinical-Genomics/oncorefiner/pull/130#discussion_r3413422492, you could also change the naming of the output file in this PR to match the current ones. IMO, if you have `subject_a_clinical.vcf.gz` it would be nice to also have a file named `subject_a_research.vcf.gz`

tests/tumor_normal.nf.test.snap
fevac: From a production perspective yes. We will always run the same steps in production so it will always be the same file name on our end.

CHANGELOG.md
```diff
@@ -91,6 +91,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#116](https://github.com/Clinical-Genomics/oncorefiner/pull/116) Updated `GENERATE_CYTOSURE_FILES` subworkflow test snapshot in sequence of [#110](https://github.com/Clinical-Genomics/oncorefiner/pull/110).
 - [#123](https://github.com/Clinical-Genomics/oncorefiner/pull/123) Changed the subworkflow test when untarring is needed in `PREPARE_REFERENCES` to check `workflow.out.vep_resources` instead of `params.outdir`
 - [#129](https://github.com/Clinical-Genomics/oncorefiner/pull/129) Changed the extra arguments `extra_args_*` to default to '' instead of null and removed empty `extra_args_snv_vep` in `test_base.config`
+- [#130](https://github.com/Clinical-Genomics/oncorefiner/pull/130) Changed output logic for `PROCESS_SNVS` to output the `research_filtered_vcf` file after annotation with vep.
```
fevac, line 94: I would argue that this is a fix and not a change. I took the liberty to add the label bug to the PR too.

fevac (approved):
Good catch! 🌟

tests/tumor_normal.nf.test.snap
fevac: If I understand correctly, this is currently not a duplicate but a filtered research file lacking annotation for CADD and VEP

CHANGELOG.md
fellen31: I've kinda viewed the "Fixed" column as should only include fixes that were fixed from the previous release, not fixes "within" a release. Would be nice to have some clear guidelines on it!

beatrizsavinhas: It is under "### `Fixed`" but do you think the wording should be different?

tests/tumor_normal.nf.test.snap
kristinebilgrav: Agree with above. But should we note down somewhere to make the name of the new `research_filtered_vcf` clearer?

CHANGELOG.md
fevac: Ah! Since it appeared as `changed` in this PR description I assumed it was the same on the changelog. Shouldn't those two be the same? 

Yeah maybe not everyone understand the same about this, it wouldn't hurt to clarify

tests/tumor_normal.nf.test.snap
beatrizsavinhas: Alright! I've noted some option we can evaluate later [here](https://github.com/Clinical-Genomics/CancerBioInfo/issues/100#issuecomment-4708273140) but for now we can move on with this implementation.

fellen31: IMO it would be nicer if this PR didn't change the name of the output file, but if it does then maybe you could explain the name change/dissappering file in the PR description and changelog?

CHANGELOG.md
beatrizsavinhas: There's some guidelines under "Types of changes" in [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) which is what is linked in the contributing guidelines in this repo.
But then I will move on with this PR and we can discuss specifics somewhere else later!

tests/tumor_normal.nf.test.snap
beatrizsavinhas: Sure, I can add more explicit information in the description and changelog! :)

beatrizsavinhas: Done now! Hope it's sufficient!

P.S.: Oj! I accidentally merged before pushing the changes to the changelog. Hope it's still OK and if not I can open a follow up PR to add this information.

PR #133, opened 2026-06-15
feat: Add pipeline stub test without `vep_plugin_files`

### LINKED ISSUE #132: Add stub test without `vep_extra_files`
### Need

As a developer,
I want add stub test without `vep_extra_files`,
because we would like to be able to run the pipeline without this file, and therefore we should have a test to make sure it can run.

### Acceptance criteria

- [ ] Pipeline stub test without `vep_extra_files` exists

fellen31:
Closes #132, first part in addressing https://github.com/Clinical-Genomics/MTP-oncoflow/issues/55.

### Added

- Added pipeline stub test without `vep_plugin_files`


tests/optional_inputs_stub.nf.test
```diff
@@ -0,0 +1,37 @@
+nextflow_pipeline {
+
+    name "Test pipeline ONCOREFINER without providing optional inputs - stub"
+    script "../main.nf"
+    tag "pipeline"
+    tag "conditional_inputs_stub"
+
+    test("No `vep_extra_files` with '-profile test' - stub") {
```
beatrizsavinhas:
```suggestion
    test("No `vep_plugin_files` with '-profile test' - stub") {
```

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#101](https://github.com/Clinical-Genomics/oncorefiner/pull/101) Added parameters for arguments in processes `PROCESS_SNVS:BCFTOOLS_VIEW_*`, `PROCESS_*:ENSEMBLVEP_VEP` and `ANNOTATE_CADD:BCFTOOLS_ANNOTATE_INDELS`.
 - [#111](https://github.com/Clinical-Genomics/oncorefiner/pull/111) Added hidden `genome_version_number` parameter, parsed by default from `params.genome`.
 - [#117](https://github.com/Clinical-Genomics/oncorefiner/pull/117) Added stub version for all pipeline and local subworkflow tests.
+- [#133](https://github.com/Clinical-Genomics/oncorefiner/pull/133) Added pipeline stub test without `vep_plugin_files`
```
beatrizsavinhas: /nit
```suggestion
- [#133](https://github.com/Clinical-Genomics/oncorefiner/pull/133) Added pipeline stub test without `vep_plugin_files` parameter.
```

beatrizsavinhas (commented):
So this looks good to me, but looking at the issue it seems that the intention is that VEP _should_ run even without providing the `vep_plugin_files` parameter. And in that case I don't really understand why adding this test would be the first thing to do in order to achieve that. Is the idea that with this test in place it will be easier in the future to see that VEP runs when the logic is changed to not depend on `params.vep_plugin_files`? 🤔

fellen31:
> So this looks good to me, but looking at the issue it seems that the intention is that VEP _should_ run even without providing the `vep_plugin_files` parameter. And in that case I don't really understand why adding this test would be the first thing to do in order to achieve that. Is the idea that with this test in place it will be easier in the future to see that VEP runs when the logic is changed to not depend on `params.vep_plugin_files`? 🤔

Yeah that was my logic about splitting the PR in two. But maybe it would have been alright just to do it in one PR.

beatrizsavinhas:
> Yeah that was my logic about splitting the PR in two. But maybe it would have been alright just to do it in one PR.

Ah OK! Then I think it makes sense to add this - which is a feature in itself - separately, but I'd say it makes more sense to add it after merging fixing the bug and then this test will instead make sure that VEP is still run.

fellen31:
> > Yeah that was my logic about splitting the PR in two. But maybe it would have been alright just to do it in one PR.
> 
> Ah OK! Then I think it makes sense to add this - which is a feature in itself - separately, but I'd say it makes more sense to add it after merging fixing the bug and then this test will instead make sure that VEP is still run.

Ok, then _ideally_ I would like to include this test in the same PR as the fix, to see that the issue has actually been fixed :smile: _But_, then I will wait this PR until the fix is merged, update the test, and ask for a re-review.

beatrizsavinhas:
> Ok, then _ideally_ I would like to include this test in the same PR as the fix, to see that the issue has actually been fixed 😄 _But_, then I will wait this PR until the fix is merged, update the test, and ask for a re-review.

I see your point! Then I go back on my previous opinion - I do think it's OK to add it in the same PR!

fellen31:
> I see your point! Then I go back on my previous opinion - I do think it's OK to add it in the same PR!

Ready for a second review (in this separate PR :smile:)!

tests/optional_inputs_stub.nf.test
kristinebilgrav, line 1: I am wondering about having the test in a separate file like this. What are the benefits of having it separate vs just adding this test to the other pipeline test files?
Also: This makes me think that we should add more optional input tests for more variables, which we should probably add an issue for?

tests/optional_inputs_stub.nf.test
```diff
@@ -0,0 +1,37 @@
...
+
+        then {
+            // stable_name: All files + folders in ${params.outdir}/ with a stable name
+            def stable_name = getAllFilesFromDir(params.outdir, relative: true, includeDir: true, ignore: ['pipeline_info/*.{html,json,txt}'])
+            // pipeline versions.yml file for multiqc from which Nextflow version is removed because we test pipelines on multiple Nextflow versions
+            def versions = removeNextflowVersion("$outputDir/pipeline_info/oncorefiner_software_mqc_versions.yml")
+            assertAll(
+                { assert workflow.success},
+                // Assert that "ENSEMBLVEP_VEP" runs when `vep_plugin_files` is not provided
+                { assert "ENSEMBLVEP_VEP" in versions.keySet()},
```
kristinebilgrav, line 57: Will this provide a failed test if not 'true'? Since I dont see what this produces in the snapshot.

kristinebilgrav (commented):
Overall looks good to me, just some questions!

tests/optional_inputs_stub.nf.test
fellen31: You could also have the `tumor_normal` and `tumor_only` tests in the same file. Together with this test if you wanted. I think having separate file is just organization. 

> Also: This makes me think that we should add more optional input tests for more variables, which we should probably add an issue for?

Yeah I think that makes sense! @beatrizsavinhas is adding one in https://github.com/Clinical-Genomics/oncorefiner/pull/103 already.

tests/optional_inputs_stub.nf.test
fellen31: Yes, this is not snapshot, it's just an assertion. The test will fail if `ENSEMBLVEP_VEP` is not in `versions`.

tests/optional_inputs_stub.nf.test
kristinebilgrav: Thats true. I think that we in the future should discuss at which level we want to separate - ex. If we want to run a non-stub optional inputs test I think those could be in the same file as this. But I dont mind this and then we can decide together at a later time. 
Okay, great, thanks!

tests/optional_inputs_stub.nf.test
kristinebilgrav: okay, great - thanks!

kristinebilgrav (approved):
🌟

PR #135, opened 2026-06-15
fix: Decouple `GENERATE_CYTOSURE_FILES` tests from pipeline configs

beatrizsavinhas:
Part of: https://github.com/Clinical-Genomics/CancerBioInfo/issues/144

### Added

- Test config for `GENERATE_CYTOSURE_FILES` with configuration for all subworkflow processes.

### Fixed

- `conf/subworkflows/generate_cytosure_files.config` so all processes regex string start with `.*:` to make sure these settings only apply for pipeline runs.
- Test for `GENERATE_CYTOSURE_FILES` so that it states the necessary parameters for the test, uses configurations from the test config and is independent from pipeline wide configuration.

### DIFF
--- a/conf/subworkflows/generate_cytosure_files.config
+++ b/conf/subworkflows/generate_cytosure_files.config
@@ -16,20 +16,20 @@
 //
 
 process {
-    withName: '.*GENERATE_CYTOSURE_FILES:.*' {
+    withName: '.*:GENERATE_CYTOSURE_FILES:.*' {
         publishDir = [
             enabled: false,
         ]
     }
 
-    withName: '.*GENERATE_CYTOSURE_FILES:FILTER_VCF' {
+    withName: '.*:GENERATE_CYTOSURE_FILES:FILTER_VCF' {
         ext.args = { [
                 '--exclude \'SVTYPE=="SGL"\'',
                 '--output-type v',
             ].join(' ') }
     }
 
-    withName: '.*GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
+    withName: '.*:GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
         ext{
             args = { [
                 "--bins 20",
--- a/subworkflows/local/generate_cytosure_files/tests/main.nf.test
+++ b/subworkflows/local/generate_cytosure_files/tests/main.nf.test
@@ -3,6 +3,7 @@ nextflow_workflow {
     name "Test Workflow GENERATE_CYTOSURE_FILES"
     script "../main.nf"
     workflow "GENERATE_CYTOSURE_FILES"
+    config "./nextflow.config"
     tag "subworkflows"
     tag "generate_cytosure_files"
 
--- a/subworkflows/local/generate_cytosure_files/tests/main.nf.test
+++ b/subworkflows/local/generate_cytosure_files/tests/main.nf.test
@@ -13,7 +13,9 @@ nextflow_workflow {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir                = "$outputDir"
+                genome_version_number = 38
+                sex                   = 'female'
             }
 
             workflow {
@@ -52,7 +54,9 @@ nextflow_workflow {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir                = "$outputDir"
+                genome_version_number = 38
+                sex                   = 'female'
             }
 
             workflow {
--- a/subworkflows/local/generate_cytosure_files/tests/main.nf.test
+++ b/subworkflows/local/generate_cytosure_files/tests/main.nf.test
@@ -13,7 +13,6 @@ nextflow_workflow {
 
         when {
             params {
-                outdir                = "$outputDir"
                 genome_version_number = 38
                 sex                   = 'female'
             }
@@ -54,7 +53,6 @@ nextflow_workflow {
 
         when {
             params {
-                outdir                = "$outputDir"
                 genome_version_number = 38
                 sex                   = 'female'
             }
--- a/subworkflows/local/generate_cytosure_files/tests/main.nf.test
+++ b/subworkflows/local/generate_cytosure_files/tests/main.nf.test
@@ -6,6 +6,8 @@ nextflow_workflow {
     tag "subworkflows"
     tag "generate_cytosure_files"
 
+    topics "versions"
+
     test("Alignment files from Tumor sample and somatic SV vcf files from TN test data") {
 
         when {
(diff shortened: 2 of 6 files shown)


conf/subworkflows/generate_cytosure_files.config
```diff
@@ -16,20 +16,20 @@
...
-    withName: '.*GENERATE_CYTOSURE_FILES:FILTER_VCF' {
+    withName: '.*:GENERATE_CYTOSURE_FILES:FILTER_VCF' {
         ext.args = { [
                 '--exclude \'SVTYPE=="SGL"\'',
                 '--output-type v',
             ].join(' ') }
     }
 
-    withName: '.*GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
+    withName: '.*:GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
```
fellen31, line 32: 💯

subworkflows/local/generate_cytosure_files/tests/nextflow.config
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    withName: '.*GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
+        ext.prefix = { "${meta2.type}_vcf2cytosure" }
+        ext.args = { [
+                "--bins 20",
+                "--sex ${params.sex}",
+                "--genome ${params.genome_version_number}"
+            ].join(' ') }
+    }
```
fellen31: There is nothing else before `GENERATE_CYTOSURE_FILES` when this test runs. 

```suggestion
    withName: 'GENERATE_CYTOSURE_FILES:FILTER_VCF' {
        ext.prefix = { "${meta.id}_filter_vcf" }
        ext.args = { [
                "--exclude 'SVTYPE==\"SGL\"'",
                '--output-type v',
            ].join(' ') }
    }

    withName: 'GENERATE_CYTOSURE_FILES:TIDDIT_COV' {
        ext.prefix = { "${meta.id}_tiddit_cov" }
    }

    withName: 'GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
        ext.prefix = { "${meta2.type}_vcf2cytosure" }
        ext.args = { [
                "--bins 20",
                "--sex ${params.sex}",
                "--genome ${params.genome_version_number}"
            ].join(' ') }
    }
```

subworkflows/local/generate_cytosure_files/tests/nextflow.config
```diff
@@ -0,0 +1,32 @@
...
+
+    withName: '.*GENERATE_CYTOSURE_FILES:TIDDIT_COV' {
+        ext.prefix = { "${meta.id}_tiddit_cov" }
+    }
+
+    withName: '.*GENERATE_CYTOSURE_FILES:VCF2CYTOSURE' {
+        ext.prefix = { "${meta2.type}_vcf2cytosure" }
+        ext.args = { [
+                "--bins 20",
+                "--sex ${params.sex}",
```
fellen31, line 27: You are not using this parameter in the test. Either use it, or hardcode/remove it.

fellen31 (changes requested):
Very nice!

kristinebilgrav (commented):
Very nice 💯

subworkflows/local/generate_cytosure_files/tests/nextflow.config
beatrizsavinhas: Do you mean that the config shouldn't rely on parameters given by the test profile and only use the ones provided in the test itself?

fellen31: If the aim is to decouple this test then no, I don't think this test should rely on configs, or test profiles used anywhere else in the pipeline.

beatrizsavinhas: Alright! Will tag you in the new changes so you can see if you agree! :)

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -12,7 +13,9 @@ nextflow_workflow {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir                = "$outputDir"
+                genome_version_number = 38
+                sex                   = 'female'
```
beatrizsavinhas, line 17: @fellen31 Do you agree with these additions as discussed in https://github.com/Clinical-Genomics/oncorefiner/pull/135#discussion_r3435412507? :)

subworkflows/local/generate_cytosure_files/tests/nextflow.config
```diff
@@ -0,0 +1,32 @@
...
+========================================================================================
+    Nextflow config file for running nf-test tests
+========================================================================================
+ Defines input files and everything required to run a fast and simple pipeline test.
+
+*/
+
+process {
+
+    withName: 'GENERATE_CYTOSURE_FILES:FILTER_VCF' {
```
kristinebilgrav:
```suggestion
    withName: 'FILTER_VCF' {
```
Suggestion: I think this would be sufficient for this config, same with the others.

kristinebilgrav (approved):
LGTM, just had one comment!

subworkflows/local/generate_cytosure_files/tests/nextflow.config
beatrizsavinhas: Indeed! Will update!

fellen31: I think the safest option is to use the process fully qualified name to always apply with the highest priority.

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -51,7 +54,9 @@ nextflow_workflow {
 
         when {
             params {
-                outdir = "$outputDir"
+                outdir                = "$outputDir"
```
fellen31:
```suggestion
```

subworkflows/local/generate_cytosure_files/tests/main.nf.test
```diff
@@ -12,7 +13,9 @@ nextflow_workflow {
...
```
fellen31:
```suggestion
```

subworkflows/local/generate_cytosure_files/tests/main.nf.test
fellen31: Sure, I don't really see why you need it, but better than before :)

fellen31 (commented):
Don't know why you need the `outdir`?

subworkflows/local/generate_cytosure_files/tests/main.nf.test
beatrizsavinhas: It's not necessary per se because `vcf2cytosure` will run anyway. But these simply match the test data - otherwise GRCh37 is the default and there is some sex default as well hardcoded in `vcf2cytosure`

beatrizsavinhas:
> Don't know why you need the `outdir`?

A missed relic of the past! Removed now, thank you! :)

Issue #75, opened 2026-04-21
Use nf-schema to validate and construct channels


mathiasbio:
### Need

As a pipeline developer,
I want to define a schema for the CSV input sheet like for the SVDB rule,
because it enforces consistent input structure and reduces downstream debugging.

### Acceptance criteria

- [ ] SVDB ch_sv_dbs CSV is validated using nf-schema
- [ ] ...any other channel that could be validated?

#### Suggested approach

Mimic Nallo approach:

- add: assets/schema_svdb_dbs.json
- Something like:

```
ch_svdb_dbs = Channel
        .fromList(samplesheetToList(ch_sv_dbs, "assets/schema_svdb_dbs.json"))
        .multiMap { row ->
            vcf_dbs:  file(row.filename)
            in_frqs:  row.in_freq_info_key
            in_occs:  row.in_allele_count_info_key
            out_frqs: row.out_freq_info_key
            out_occs: row.out_allele_count_info_key
        }
```

#### Other notes

See comment from @fellen31 : https://github.com/Clinical-Genomics/oncorefiner/pull/73#discussion_r3116222814

Example schema from nallo: https://github.com/genomic-medicine-sweden/nallo/blob/5a9ad3ae07a7027bd4d271bd46699493c62fac79/assets/svdb_query_vcf_schema.json

fevac:
@mathiasbio please add the description here, if unclear ask @fellen31 for input

PR #136, opened 2026-06-15
feat: Add nf-schema validation to SVDB database inputs

fellen31:
At least partly addresses https://github.com/Clinical-Genomics/oncorefiner/issues/75

### Added

-  Added nf-schema validation to SVDB database inputs (`svdb_query_dbs`)

PR #137, opened 2026-06-15
feat: add nf-schema validation of VEP and fixed VEP able to run without `vep_plugin_files`

fellen31:
Adresses second part of https://github.com/Clinical-Genomics/oncorefiner/issues/75, and https://github.com/Clinical-Genomics/MTP-oncoflow/issues/55. 

### Added

- Added nf-schema validation of VEP files rather than custom CSV parsing

### Changed
- `vep_plugin_files` help text to mention that now CSV/TSV/JSON/YAML can be used. 

### Fixed
- Fixed VEP able to run without `vep_plugin_files`


kristinebilgrav (approved):
Very nice! 💯

fellen31:
> Very nice! 💯

Actually, I realized that I accidentally also fixed https://github.com/Clinical-Genomics/MTP-oncoflow/issues/55 in this issue. Will update the changelog, add one more edit to `main.nf`, and ask for a re-review if that's okay.

main.nf
```diff
@@ -108,21 +109,8 @@ workflow CLINICALGENOMICS_ONCOREFINER {
...
-                } else {
-                    error("\nVep database file ${f} does not exist.")
-                }
-            }
-            .collect()
-            .set {ch_vep_extra_files}
-    }
-
+    ch_vep_extra_files = val_vep_plugin_files ? channel.fromList(samplesheetToList(val_vep_plugin_files, 'assets/vep_plugin_files_schema.json')).collect()
+                                              : channel.value([])
```
beatrizsavinhas, line 113: Would it be worth it to have custom channel factory for this?
Something like `validatedChannelFromCSV(<csv>, <schema>, <fallback?>)` but perhaps there's a more general name that we can use here instead of "CSV" - tabular file? 🤔

CHANGELOG.md
```diff
@@ -93,6 +94,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#129](https://github.com/Clinical-Genomics/oncorefiner/pull/129) Changed the extra arguments `extra_args_*` to default to '' instead of null and removed empty `extra_args_snv_vep` in `test_base.config`
 - [#130](https://github.com/Clinical-Genomics/oncorefiner/pull/130) Changed output logic for `PROCESS_SNVS` to output the `research_filtered_vcf` file after annotation with vep.
 - [#131](https://github.com/Clinical-Genomics/oncorefiner/pull/131) Fixed strict Nextflow syntax compatibility in `ANNOTATE_CADD`
+- [#137](https://github.com/Clinical-Genomics/oncorefiner/pull/134) Fixed VEP able to run without `vep_plugin_files`
```
beatrizsavinhas: /nit
```suggestion
- [#137](https://github.com/Clinical-Genomics/oncorefiner/pull/134) Fixed VEP able to run without `vep_plugin_files`  parameter.
```

beatrizsavinhas (commented):
Sorry, I'm thinking I've missed something. Could you explain how this fixes the issue of VEP not starting without `vep_plugin_files`? Is it due to the fallback channel no longer being empty but `channel.value([])`? 🤔

nextflow_schema.json
```diff
@@ -174,8 +174,8 @@
                     "format": "file-path",
                     "description": "Databases used by both named and custom plugins to annotate variants.",
                     "fa_icon": "fas fa-file-csv",
-                    "help_text": "Path to a file containing the absolute paths to databases and their indices used by VEP's custom and named plugins resources defined within the vcfanno toml file. One line per resource.",
-                    "pattern": "^\\S+\\.csv$",
+                    "help_text": "Path to a CSV/TSV/JSON/YAML file with vep_files as header, and then the absolute paths to databases and their indices used by VEP's custom and named plugins resources defined within the vcfanno toml file.",
+                    "pattern": "^\\S+\\.(csv|tsv|json|ya?ml)$",
```
beatrizsavinhas, line 178: Would be nice to include these changes in the PR description even if you think they are not important enough to be mentioned in the changelog.

fellen31: Sure, added.

main.nf
fellen31: Yeah, I think we could add one in the future!

fellen31:
> Sorry, I'm thinking I've missed something. Could you explain how this fixes the issue of VEP not starting without `vep_plugin_files`? Is it due to the fallback channel no longer being empty but `channel.value([])`? 🤔

Yeah, exactly! Previously it was an empty channel so the process doesn't start, but now it's a channel with an empty list, so it still runs.

main.nf
fellen31: Think it would be easiest after both this and https://github.com/Clinical-Genomics/oncorefiner/pull/136 is merged. I started that PR before knowing that we should also use nf-schema for the VEP file.

nextflow_schema.json
beatrizsavinhas, line 178: Actually, still about this, in my opinion this PR adds support for `vep_plugin_files` to be of the CSV/TSV/JSON/YAML format whereas before only CSV was supported, which is a new feature that should be mentioned under "Added" in the changelog.

nextflow_schema.json
```diff
@@ -174,8 +174,8 @@
...
                     "mimetype": "text/csv",
                     "schema": "assets/vep_plugin_files_schema.json"
```
beatrizsavinhas, line 180: I wonder if there is a way to somehow provide this `schema` value automatically for channel generation above. Would be neat instead of having to repeat the schema file path! 🤔

beatrizsavinhas (approved):
> Yeah, exactly! Previously it was an empty channel so the process doesn't start, but now it's a channel with an empty list, so it still runs.

Makes sense!

Looks good to me! 🌟

nextflow_schema.json
fellen31: Ok, added!

nextflow_schema.json
fellen31: Yeah, definitely!

PR #140, opened 2026-06-18
feat: Add linx combine and vcf annotation local modules

kristinebilgrav:
Main issue: https://github.com/Clinical-Genomics/MTP-oncoflow/issues/67

### Added

- local modules: `combine_linx_tsv` and `annotate_vcf_by_id` , which will combine LINX tsv file into one, and annotate a VCF using a TSV and the ID. 
    -  local module test and snapshot

Note: 
Scripts added and reviewed here: https://github.com/Clinical-Genomics/oncorefiner/pull/114

Main branch: 
https://github.com/Clinical-Genomics/oncorefiner/pull/115


modules/local/annotate_vcf_by_id/environment.yml
kristinebilgrav, line 1: Is this file needed?

modules/local/annotate_vcf_by_id/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        'https://wave.seqera.io/view/inspect?image=community.wave.seqera.io/library/pysam_click_python:ff252e0bf843a2dd&platform=linux/amd64' :
+        'community.wave.seqera.io/library/pysam_click_python:ff252e0bf843a2dd' }"
+
+    input:
+    tuple val(meta), path(vcf_file), path(tsv_file), path(header_file)
+
+    output:
+    tuple val(meta), path("${prefix}.vcf.gz"), emit: vcf
+    tuple val(meta), path("${prefix}.vcf.gz.tbi"), emit: tbi
+    //tuple val("${task.process}"), val('annotate_vcf_by_id'), topic: versions, emit: versions_annotate_vcf //not working?
```
kristinebilgrav: Could need some help here. Although I dont see this as a very large must-have currently

modules/local/combine_linx_tsv/tests/main.nf.test
```diff
@@ -0,0 +1,35 @@
...
+
+    name "Test Process COMBINE_LINX_TSV"
+    script "../main.nf"
+    process "COMBINE_LINX_TSV"
+    tag "modules"
+    tag "modules_local"
+    tag "combine_linx_tsv"
+    config "./nextflow.config"
+
+    test("Combine LINX TSVs, tumor-normal") {
```
kristinebilgrav: Using TN data. The test would be identical for TO data, so would argue its not needed for now.

modules/local/annotate_vcf_by_id/main.nf
```diff
@@ -0,0 +1,38 @@
+process ANNOTATE_VCF_BY_ID{
+    tag "$meta.id"
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
```
kristinebilgrav, line 5: Made these with: https://seqera.io/containers/
Hope that is sufficient and stable?

modules/local/annotate_vcf_by_id/main.nf
beatrizsavinhas: The script has no functionality to return its version. You can either add that or hardcode the version here!

modules/local/annotate_vcf_by_id/main.nf
beatrizsavinhas: Would assume so, but I have no experience with this so perhaps someone else can help here!

modules/local/annotate_vcf_by_id/environment.yml
beatrizsavinhas: Looks like it is a necessary file for a module - see https://nf-co.re/docs/developing/tutorials/writing-nf-core-modules/5-writing-modules#the-environmentyml-file.

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,56 @@
...
+
+    test("Annotate VCF by ID, tumor-normal") {
+
+        setup {
+            run("COMBINE_LINX_TSV") {
+                script "modules/local/combine_linx_tsv/main.nf"
+                process {
+                    """
+                    input[0] = [
+                    [ id:'test' ],
```
beatrizsavinhas: Would recommend following the updated meta structure! Same for all other inputs.

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,56 @@
...
+
+            process {
+                """
+                ch_input = channel.of([
+                    [ id:'test' ],
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.sv.vcf.gz', checkIfExists: true)
+                    ])
+                    .join(COMBINE_LINX_TSV.out.tsv, failOnMismatch: true)
+                    .combine( channel.fromPath('$projectDir/assets/sv_annotation_header.txt') )
+                input[0] = ch_input.map { meta, vcf, tsv, header -> tuple(meta, vcf, tsv, header) }
```
beatrizsavinhas: Suggest moving and modifying the code above to line 40 so that the channel operations already return the input with the desired structure.

modules/local/annotate_vcf_by_id/tests/main.nf.test
beatrizsavinhas, line 1: Add stub test.

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,56 @@
...
+
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    process.out.vcf.collect { meta, vcf -> [meta, file(vcf).name, ":variantsMD5,", path(vcf).vcf.variantsMD5 ]} ,
```
beatrizsavinhas: Correct this logic so the snapshot has the correct structure:  `"<file_name> :variantsMD5, <md5sum>"`

modules/local/combine_linx_tsv/tests/nextflow.config
```diff
@@ -0,0 +1,7 @@
+process {
+
+    withName: 'COMBINE_LINX_TSV' {
+        ext.prefix = { "${meta.id}_linxcombined" }
+
```
beatrizsavinhas:
```suggestion
    withName: 'COMBINE_LINX_TSV' {
        ext.prefix = { "${meta.id}_linxcombined" }
```

For consistent syntax/spacing in both configs.

modules/local/combine_linx_tsv/main.nf
```diff
@@ -0,0 +1,29 @@
...
+    output:
+    tuple val(meta), path("*.tsv"), emit: tsv
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    combine_linx_tsv.py -f ${fusion_tsv_file} -b ${breakends_tsv_file} -sv ${sv_tsv_file} -o ${prefix}.tsv
```
beatrizsavinhas: Separate into different line to keep consistent syntax.

beatrizsavinhas (changes requested):
Leaving an initial review as I think there are some aspects that still need work!

`meta.yml` is missing for both modules and `environment.yml` for one of them. See https://nf-co.re/docs/developing/tutorials/writing-nf-core-modules/5-writing-modules for more information about writing modules.

Stub tests are also missing for both modules.

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: Think I did this now!

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: Which meta structure do you mean? These are modules, so they can be anything no?

beatrizsavinhas: Yes, the test for the modules can be completely independent from the pipeline. But since this is a local module now being built from scratch, I thought it makes sense to keep the meta consistent with what the channels will actually look like. We don't have to go back and update existing tests, but we might as well code the new ones better from the get-go.

modules/local/combine_linx_tsv/tests/main.nf.test
beatrizsavinhas: Would recommend improving the name of the test though, to be more explanatory. "Combine LINX TSVs" is already explicit from the overarching test name.

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    process.out.vcf.collect { meta, vcf -> [meta, file(vcf).name + ":variantsMD5," + path(vcf).vcf.variantsMD5 ]} ,
+                    process.out.tbi.collect { meta, tbi -> [meta, file(tbi).name ]} ,
+                    process.out.findAll { key, val -> key.startsWith("versions") } ,
```
beatrizsavinhas: This can be done with the new syntax which you already implemented in a recent PR!

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,107 @@
...
+
+            process {
+                """
+                ch_input = channel.of([
+                    [ id:'test' ],
+                    file(params.pipelines_testdata_base_path + 'testdata/GRCh38/tumor_normal/subject_a.tumor.purple.sv.vcf.gz', checkIfExists: true)
+                    ])
+                    .join(COMBINE_LINX_TSV.out.tsv, failOnMismatch: true)
+                    .combine( channel.fromPath('$projectDir/assets/sv_annotation_header.txt') )
+                    .map { meta, vcf, tsv, header -> tuple(meta, vcf, tsv, header) }
```
beatrizsavinhas: Isn't this just creating a channel with the exact same structure?
```suggestion
```

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,107 @@
...
+                input[0] = ch_input
```
beatrizsavinhas: Same comment as before. See input below too.

modules/local/annotate_vcf_by_id/environment.yml
```diff
@@ -0,0 +1,7 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - conda-forge::python=3.14.3
+  - conda-forge::pysam=0.24.0
+  - conda-forge::click=8.2.1
```
beatrizsavinhas, line 7: Isn't `csv` also used in this script? Does it not need to be specified as a dependency?

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,107 @@
...
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    process.out.vcf.collect { meta, vcf -> [meta, file(vcf).name ]} ,
+                    process.out.tbi.collect { meta, tbi -> [meta, file(tbi).name ]} ,
```
beatrizsavinhas: Special handling not needed for the stub test since you expect and want to test that the files are empty.
```suggestion
                    process.out
```

modules/local/annotate_vcf_by_id/main.nf
```diff
@@ -0,0 +1,38 @@
...
+        'https://wave.seqera.io/view/inspect?image=community.wave.seqera.io/library/pysam_click_python:ff252e0bf843a2dd&platform=linux/amd64' :
+        'community.wave.seqera.io/library/pysam_click_python:ff252e0bf843a2dd' }"
+
+    input:
+    tuple val(meta), path(vcf_file), path(tsv_file), path(header_file)
+
+    output:
+    tuple val(meta), path("${prefix}.vcf.gz"), emit: vcf
+    tuple val(meta), path("${prefix}.vcf.gz.tbi"), emit: tbi
+    tuple val("${task.process}"), val('annotate_vcf_by_id'), val('1.0'), topic: versions, emit: versions_annotate_vcf
```
beatrizsavinhas, line 14: This seems to have a different structure than the standard:
```shell
"<task.process>": {
                    "<tool>": "<version>"
                },
```

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+  - genomics
+tools:
+  - "python":
+      description: "Annotates VCF files by ID."
+      homepage: "https://www.python.org/"
+      documentation: "https://docs.python.org/3/"
+      tool_dev_url: "https://github.com/python/cpython"
+      doi: ""
+      licence: ["MIT"]
+      identifier: biotools:python
```
beatrizsavinhas: The script is the tool that this module calls, not python.

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
```
beatrizsavinhas: Is the input channel meta supposed to include sample information? Looks like the meta will come from the vcf channel, which is a case file.
Also give an example that matches the test.
Same below.

modules/local/combine_linx_tsv/main.nf
```diff
@@ -0,0 +1,34 @@
...
+    tuple val(meta), path("*.tsv"), emit: tsv
+    tuple val("${task.process}"), val('combine_linx_tsv'), val('1.0'), topic: versions, emit: versions_combine_linx_tsv
+
+    when:
+    task.ext.when == null || task.ext.when
+
+    script:
+    def prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    combine_linx_tsv.py \
```
beatrizsavinhas: Correct syntax for all lines.
```suggestion
    combine_linx_tsv.py \\
```

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: Seems like it!

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: Ah thats true. I followed what the nf-core modules were doing, but can do it like before instead.

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: For some reason when I change to this it snapshots local paths, so would need help to fix that in that case  👀

modules/local/annotate_vcf_by_id/environment.yml
kristinebilgrav: Its a built-in module, do I also need to specify that then?

modules/local/annotate_vcf_by_id/main.nf
kristinebilgrav: Hmm, when looking at other modules, its written the same way. Could you give example of how to fix it?

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: Okay, added them now!

modules/local/annotate_vcf_by_id/main.nf
beatrizsavinhas: Yeah, I looked at a snapshot from another module and indeed it looks like the one here! So you can disregard this comment.
There must be some kind of parsing/modification done when printing these to the `*_software_mqc_versions.yml` file that makes them have the format I mentioned above.

modules/local/combine_linx_tsv/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+                """
+            }
+        }
+
+        then {
+            assert process.success
+            assertAll(
+                { assert snapshot(
+                    process.out.tsv ,
+                    process.out.findAll { key, val -> key.startsWith("versions") } ,
```
beatrizsavinhas:
```suggestion
                    process.out.tsv,
                    process.out.findAll { key, val -> key.startsWith("versions") },
```
Looks like there are some trailing whitespaces here. You can try to run `nf-core modules lint` on these modules to see if there are any corrections that need to be done.

modules/local/annotate_vcf_by_id/tests/main.nf.test
beatrizsavinhas: Testing it locally it seems that the `tbi` output works as intended, but the `vcf` is indeed snapshot with full paths. @fellen31 do you have any idea of why this might be? 🤔

modules/local/annotate_vcf_by_id/main.nf
fellen31, line 14: Would add a warning that you need to change this manually when updating the script. It is notoriously difficult to remember: https://github.com/Anifowak/modules/blob/22e0f3ed5fe33b3190b793d388d74cb6b9398061/modules/nf-core/bioawk/main.nf#L16

modules/local/annotate_vcf_by_id/main.nf
fellen31: That's okay.

modules/local/annotate_vcf_by_id/main.nf
```diff
@@ -0,0 +1,38 @@
+process ANNOTATE_VCF_BY_ID{
+    tag "$meta.id"
+
+    conda "${moduleDir}/environment.yml"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://wave.seqera.io/view/inspect?image=community.wave.seqera.io/library/pysam_click_python:ff252e0bf843a2dd&platform=linux/amd64' :
```
fellen31: This doesn't look correct to me. Should be something like: `https://community-cr-prod.seqera.io/docker/registry/v2/blobs/sha256/90/9099ea4235b8fe50ac8f4863d30f04469fac122e7b394474c64abe8960662b2c/data`.

modules/local/annotate_vcf_by_id/tests/nextflow.config
fellen31, line 1: I would delete this, I don't really see why it's necessary. You only have one output file. If you want to change the output name you could modify the meta.id in the test.

modules/local/combine_linx_tsv/tests/nextflow.config
fellen31, line 1: I would delete this, I don't really see why it's necessary. You only have one output file.

modules/local/annotate_vcf_by_id/tests/main.nf.test
fellen31: You need to use echo | gzip > file.vcf.gz  https://nf-co.re/docs/specifications/components/modules/general#stub-gzip-files-must-use-echo-and-pipe)

modules/local/annotate_vcf_by_id/main.nf
fellen31: Screenshot 2026-06-24 at 16 20 11

modules/local/annotate_vcf_by_id/main.nf
kristinebilgrav: Thats true - thanks!

kristinebilgrav: Okay, thanks!

modules/local/annotate_vcf_by_id/tests/main.nf.test
kristinebilgrav: That solved it - thanks!!

modules/local/annotate_vcf_by_id/tests/nextflow.config
kristinebilgrav: Okay, will remove!

modules/local/combine_linx_tsv/tests/nextflow.config
kristinebilgrav: Removing

modules/local/annotate_vcf_by_id/main.nf
kristinebilgrav: Updated now, thanks!

modules/local/annotate_vcf_by_id/environment.yml
beatrizsavinhas: Ah, in that case i don't think so!

modules/local/annotate_vcf_by_id/main.nf
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    stub:
+    prefix = task.ext.prefix ?: "${meta.id}"
+    """
+    echo | gzip >  ${prefix}.vcf.gz
+    touch ${prefix}.vcf.gz.tbi
+    """
+
+
```
beatrizsavinhas:
```suggestion
```

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+  - genomics
+tools:
+  - "python":
+      description: "Annotates VCF files by ID."
+      homepage: "https://www.python.org/"
+      documentation: "https://docs.python.org/3/"
+      tool_dev_url: "https://github.com/python/cpython"
+      doi: ""
+      licence: ["MIT"]
+      identifier: annotate_vcf_by_id
```
beatrizsavinhas: See earlier comment - https://github.com/Clinical-Genomics/oncorefiner/pull/140#discussion_r3466035515.

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing case information
+          e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor_normal', sex: 'female']`
```
beatrizsavinhas: This meta contains sample information and does not adhere to the structure we have for `metadata_case_file`.

modules/local/annotate_vcf_by_id/tests/main.nf.test
```diff
@@ -0,0 +1,101 @@
...
+
+    test("Annotate VCF by ID, tumor-normal") {
+
+        setup {
+            run("COMBINE_LINX_TSV") {
+                script "modules/local/combine_linx_tsv/main.nf"
+                process {
+                    """
+                    input[0] = [
+                    [id: 'test_linxannotated', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor_normal', sex: 'female'],
```
beatrizsavinhas: This metadata seems to have the incorrect format as mentioned below.

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+    - vcf_file:
+        type: file
+        description: VCF file to be annotated
+        pattern: "*.vcf"
+    - tsv_file:
+        type: file
+        description: TSV file to use for annotation with the format ID\tTAG1\tTAG2\t...\tTAGN
```
beatrizsavinhas: Does the files require a header?

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+        type: file
+        description: VCF header file
+        pattern: "*.txt"
+output:
+  - vcf:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing case information
+            e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor_normal', sex: 'female']`
```
beatrizsavinhas: Same as above.

modules/local/annotate_vcf_by_id/meta.yml
```diff
@@ -0,0 +1,58 @@
...
+      - "*.vcf":
+          type: file
+          description: Annotated VCF file
+          pattern: "*.vcf"
+  - tbi:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing case information
+            e.g. `[id: 'test', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor_normal', sex: 'female']`
```
beatrizsavinhas: Same as above.

modules/local/combine_linx_tsv/tests/main.nf.test
```diff
@@ -0,0 +1,65 @@
...
+
+    topics "versions"
+
+    test("Test combining the LINX TSVs into annotation compatible TSV, tumor-normal") {
+        when {
+
+            process {
+                """
+                input[0] = [
+                    [id: 'test_linxcombined', case_id: 'test', sample_id: 'subject_a', sample_type: 'tumor_normal', sex: 'female'],
```
beatrizsavinhas: Same as above.

modules/local/combine_linx_tsv/meta.yml
```diff
@@ -0,0 +1,48 @@
...
+      tool_dev_url: "https://github.com/python/cpython"
+      doi: ""
+      licence: ["MIT"]
+      identifier: biotools:python
+input:
+  - - meta:
+        type: map
+        description: |
+          Groovy Map containing sample information
+          e.g. `[ id:'sample1', single_end:false ]`
```
beatrizsavinhas: Incorrect information.

modules/local/combine_linx_tsv/meta.yml
```diff
@@ -0,0 +1,48 @@
...
+        type: file
+        description: LINX SV TSV file
+        pattern: "*.tsv"
+output:
+  - tsv:
+      - meta:
+          type: map
+          description: |
+            Groovy Map containing sample information
+            e.g. `[ id:'sample1', single_end:false ]`
```
beatrizsavinhas: Same as above.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genmod_score_config` parameter.
 - [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `genomic-medicine-sweden/vcf_annotate_score_genmod` subworkflow to `PROCESS_SNVS`.
 - [#103](https://github.com/Clinical-Genomics/oncorefiner/pull/103) Added `tests/optional_inputs_stub.nf.test` to test running the pipeline without optional inputs. Includes stub test for running the pipeline without providing the `genmod_score_config` parameter.
+- [#140](https://github.com/Clinical-Genomics/oncorefiner/pull/140) Added local modules along with tests `COMBINE_LINX_TSV` and `ANNOTATE_VCF_BY_ID` for combining LINX TSV files into one and annotating a VCF by using the VCF ID and a TSV annotation file.
```
beatrizsavinhas: Suggest separating in one entry per module and make the explanation more clear.

beatrizsavinhas (approved):
Left some comments on small details and incorrect meta structure and information. Otherwise it looks good to me!

modules/local/annotate_vcf_by_id/meta.yml
kristinebilgrav: Yes

beatrizsavinhas: What I meant is that in that case, the information should be added here!

kristinebilgrav: I added `header required` before merging, dont know why it does not appear here, but you can see it in the feature branch here: https://github.com/Clinical-Genomics/oncorefiner/pull/115/changes#diff-b140b6bfe7baa0c2aa4d21df101565bba68fe0a28fc6fa1ed8831fca11888d13

PR #141, opened 2026-06-18
refactor: `GENERATE_CYTOSURE_FILES` to emit workflow outputs

beatrizsavinhas:
Part of: https://github.com/Clinical-Genomics/CancerBioInfo/issues/144

### Added

- `emit` statement so the subworkflow emits `VCF2CYTOSURE.out.cgh`.

### Changed

- Updated test to snapshot workflow output and versions, instead of published output directories and files as was previously done.


fellen31 (approved):
💯

kristinebilgrav (approved):
Nice, much clearer! 🧹

PR #145, opened 2026-06-22
test: small fixes to `GENERATE_CYTOSURE_FILES` tests from #135

beatrizsavinhas:
Fixes to https://github.com/Clinical-Genomics/oncorefiner/issues/135.

### Changed

- Config regex to include the fully qualified name of the process in order for the settings to be applied with the highest priority.

- Remove param dependency from each test and move the values to the config file instead.


beatrizsavinhas:
My thought was that a changelog entry is not relevant since this PR doesn't actually change any logic but just includes some choices that would have otherwise been included in https://github.com/Clinical-Genomics/oncorefiner/pull/135. Let me know if you don't agree!

fellen31:
> My thought was that a changelog entry is not relevant since this PR doesn't actually change any logic but just includes some choices that would have otherwise been included in #135. Let me know if you don't agree!

Can you add this PR number to the same changelog entry as #135?

beatrizsavinhas:
> Can you add this PR number to the same changelog entry as #135? 

Naaah don't want to _manipulate_ the log like that! 😅 I'll just add another entry that mentions #135 then!

kristinebilgrav (approved):
👍

PR #152, opened 2026-06-24
chore: Migrate `prepare_references` to workflow outputs

fellen31:
Closes https://github.com/Clinical-Genomics/MTP-oncoflow/issues/83.

### Removed

- Removed publishDir directive for `PREPARE_REFERENCES`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,3 +128,4 @@ Initial release of Clinical-Genomics/oncorefiner, created with the [nf-core](htt
 - [#51](https://github.com/Clinical-Genomics/oncorefiner/pull/51) Removed unused parameter `custom_extra_files`.
 - [#125](https://github.com/Clinical-Genomics/oncorefiner/pull/125) Removed `input` parameter and samplesheet logic and documentation.
 - [#139](https://github.com/Clinical-Genomics/oncorefiner/pull/139) Removed SVDB output file, since it's an intermediary file not useful to have in the outputs.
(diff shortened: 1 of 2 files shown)


fevac (approved):
🚀

kristinebilgrav (approved):
Great! 🌟  Expected more changes, but I guess the wf already emits what is needed and we dont need the output published to a directory, so this is all that is needed?

beatrizsavinhas (approved):
Greeeeat! 🌟 

Just clean up the PR description for the sake of a clean commit message! 🧹 

Also, I see that the CI action to run nf-test didn't actually run. I don't anticipate this to change anything in the output structure, but we've seen in the past that after merging a PR that passed all checks there were failing tests that were only discovered in subsequent PRs. Anyway, I don't think we need to do anything about this here/now but just commenting to keep this in mind.

fellen31:
> Great! 🌟 Expected more changes, but I guess the wf already emits what is needed and we dont need the output published to a directory, so this is all that is needed?

Yeah I think this is the only change needed. 

> Also, I see that the CI action to run nf-test didn't actually run. I don't anticipate this to change anything in the output structure, but we've seen in the past that after merging a PR that passed all checks there were failing tests that were only discovered in subsequent PRs. Anyway, I don't think we need to do anything about this here/now but just commenting to keep this in mind.

Yeah, this is a bit problematic. You could add `conf/*` to the nf-test trigger list to circumvent, i..e. https://github.com/genomic-medicine-sweden/nallo/blob/9bf6a2a1d1e62f6fafc01e483dee19e0954f6206/nf-test.config#L28.

